PR #2, opened 2023-01-25
LW-R8: Empty Author Query appears for the author user (LEAN-2001)

asouqi:


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@manuscripts/transform",
   "description": "ProseMirror transformer for Manuscripts applications",
-  "version": "1.0.1",
+  "version": "1.0.1-LEAN-2001",
```
hhusban, line 4: shouldn't this be changed to a regular versioning

asouqi: so, there is no need for testing that first. before merging it.
CC: @mnatsheh

mnatsheh:
@asouqi @hhusban what's the status of this? Sorry I missed the previous comment.

asouqi:
> what's the status of this? Sorry I missed the previous comment.

oh nevermind I will close this PR. as we fix this issue on the comment rework [PR](https://github.com/Atypon-OpenSource/manuscripts-transform/pull/4)

PR #4, opened 2023-02-08
LEAN-2105 -- moving comments to a separate node instead of local attr

hhusban:


asouqi (approved):
and this should work without any change in the FE?

src/transformer/__tests__/__snapshots__/decode.test.ts.snap
```diff
@@ -5744,6 +5714,47 @@ Object {
       ],
       "type": "section",
     },
+    Object {
+      "attrs": Object {
+        "id": "undefined:9158450B-F1AF-44B7-B884-4EDC4AA27107",
```
mbartenev-atypon, line 5719: Is that ok that is it "undefined"?

hhusban: Oh, since there is no ObjectType associated with this node the id would look like this, it won't break anmything, but yeah it doesn't look good

hhusban: would it be better to remove the whole attr section (this object is just a placeholder, it will not be represented in the modelMap)

src/transformer/decode.ts
```diff
@@ -285,7 +293,7 @@ export class Decoder {
           ),
           attribution: model.attribution,
           alternatives: model.alternatives,
-          comments: this.createCommentsNode(model),
+          comments: [...commentNodes.map((c) => c.attrs.id).values()],
```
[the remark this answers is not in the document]
hhusban: I implemented this thinking it was a map rather than an array.

mbartenev-atypon (approved):

src/schema/index.ts
```diff
@@ -76,6 +76,7 @@ import { text } from './nodes/text'
 import { tocElement } from './nodes/toc_element'
 import { tocSection } from './nodes/toc_section'
 import { Marks, Nodes } from './types'
+import {comment_list} from "./nodes/comment_list";
```
mbartenev-atypon, line 79: One more thing, seems like we use camelCase. Also please check linting. There are no spaces around the comment_list

mbartenev-atypon (changes requested):

mbartenev-atypon (approved):

asouqi:
toDOM will be optional in this case:
- We have a custom view node, that we add in `nodeViews` so it will assume we are going to return an HTML element for the mapped node.
- Inline content text in this case no need for it.
I tried to add a custom view for `comment_list` but we need to return a dom element with attrs contenteditable. so we should return something

The only two solutions I can see:
1.  add toDOM for `cmment_list` to be like this:
     `toDOM: () => ['meta']` and the only side effect for this is that we see in the article body meta node
2. the optimal solution is to move `comment_list` to the doc node so it will be like this `content: 'manuscript comment_list*'` and when we build the editor will read just the manuscript node. but I don't know if this will introduce other issues!

mbartenev-atypon:
> 

I think what we need here is to create a custom view for the meta section (parent element) and omit contentDOM property. That way prosemirror will not attempt to render any of it children.

asouqi:
>I think what we need here is to create a custom view for the meta section (parent element) and omit contentDOM property in it. That way prosemirror will not attempt to render any of its children.

there is no need for a custom view, as we return just the dom node without content.
> number zero (“hole”) in it to indicate where the node's content should be inserted.
https://prosemirror.net/docs/ref/#model.NodeSpec.toDOM

mbartenev-atypon:
If you see that as fitting solution, I don't mind. I think I was talking about a different approach.

PR #9, opened 2023-03-21
Use prosemirror document for showing and saving comments (LEAN-2322)

asouqi:


hhusban (approved):
Looks good

src/schema/nodes/comment.ts
```diff
@@ -35,5 +39,8 @@ export const comment: NodeSpec = {
     contents: { default: '' },
     target: { default: '' },
     selector: { default: null },
+    resolved: { default: null },
+    contributions: { default: null },
+    originalText: { default: null },
```
mbartenev-atypon, line 44: If it's possible it would be advisable to use default values corresponding to the actual type of the values. I'm guessing originalText would be empty string - '', resolved: false, contributions - empty array (not sure)

mbartenev-atypon (approved):

PR #14, opened 2023-04-10
LEAN-2066 -- Add comments to keyword: BE Implementation

hhusban:


asouqi (changes requested):
can we update the test case data with a comment for the keyword, like this one `encode highlight markers` in `encode.test.ts`

mnatsheh (commented):
Is it possible to move the author queries outside the individual keywords instead of this? it feels that we're hacking the commenting logic by making keywords highlightable even though they're not.

mnatsheh (approved):
I was thinking that this would happen at JATS processing time instead of here, but this is fine too.

asouqi (approved):

src/transformer/__tests__/__helpers__/highlights.ts
```diff
@@ -289,6 +290,79 @@ export const createTestModelMapWithKeywords = () => {
...
+  modelMap.set(keyword._id, keyword)
+
+  manuscript.keywordIDs = [keyword._id]
+
+  modelMap.set(manuscript._id, manuscript)
+
+  const comment: CommentAnnotation = {
+    _id: 'MPCommentAnnotation:3AAEE869-7DC9-4392-B8C6-9C5FCA522120',
+    objectType: 'MPCommentAnnotation',
+    target: 'MPKeywordGroup:test',
```
asouqi, line 336: this should be section ID.
if you looked at the keyword it will be like this:
```
{
            "type": "keywords_section",
            "attrs": {
                "id": "MPSection:1EA32707-96E5-4CFF-856E-5AD75D814295",
                "dataTracked": null
            },
```

asouqi (changes requested):

asouqi (changes requested):
target now appears to me like this in the modelMap:
```
    "target": "00af859f-074e-499a-8c3e-0c6c1a2c7685",
```

and I can't see the keyword comment in the `comment_list` as we have it wrong in the modelMap.

if we can update the JATS article we use for testing to add keywords with authorQuery, that will show the issue.
```
<kwd-group kwd-group-type="author"><title>KEYWORDS: </title><kwd>suicide attempters<?AuthorQuery id="ntQ1" queryProcess="QA Test" queryText="suicide attempters looks like a strange keyword"?></kwd><kwd>schizophrenia</kwd><kwd><italic>D&#x03B2;H<?AuthorQuery id="ntQ2" queryProcess="qa Test" queryText="check the special characters for the keywords"?></italic></kwd><kwd>attention</kwd><kwd>polymorphism</kwd></kwd-group>

```

asouqi (approved):

PR #15, opened 2023-04-18
LEAN-2402 -- Remove bundled data

hhusban:


asouqi (changes requested):
can we remove: `loadBundledData` & [shared-data](https://github.com/Atypon-OpenSource/manuscripts-transform/blob/955b92026f2413b2720865da5213b2fc6ee3c11b/src/transformer/shared-data.ts)

hhusban:
@asouqi I update the PR and removed both loadBundledData & shared-data, plus any lingering functions left

asouqi (approved):

mnatsheh (approved):

mnatsheh:
Do we still need the manuscripts/data dependency?

hhusban:
@mnatsheh  there is still a couple of references under, which are used for in the front-parser. 
src/transformer/bundles-data.ts

PR #16, opened 2023-04-18
LEAN-2410 -- SSec-100 - Conflict of Interest Statements - Implementation

hhusban:


asouqi (approved):

src/jats/jats-exporter.ts
```diff
@@ -2158,7 +2162,8 @@ export class JATSExporter {
 
   sectionToFootnote = (section: Element, fnType: string) => {
     const footNote = this.document.createElement('fn')
-    footNote.setAttribute('fn-type', fnType)
+    const footnoteType = chooseJatsFnType(fnType)
+    footNote.setAttribute('fn-type', footnoteType)
```
mnatsheh, line 2166: This should happen in the serializer.

mnatsheh (changes requested):
I don't like moving the data from the back to the front. I think we should add a new `moveSectionsToFront` similar to `moveSectionsToBack` that looks for the appropriate section, and creates the author notes fn. `moveSectionsToBack` will have to be updated to not include that particular fn-type.

src/jats/jats-exporter.ts
hhusban: I processed the footnotes here since this section soul purpose is to process the footnotes.

Re- (I think we should add a new moveSectionsToFront similar to moveSectionsToBack ...)
I wanted to, but it seemed like it gonna add more complexity to changes.

PR #18, opened 2023-04-30
LEAN-2403: Remove bundle detection from JATS import

mbetamony:

PR #20, opened 2023-05-02
LEAN-2407

asouqi:


mnatsheh:
There are some failing tests.

asouqi:
> There are some failing tests.

fix

package.json
```diff
@@ -30,6 +30,7 @@
   },
   "dependencies": {
     "@manuscripts/json-schema": "^2.0.0",
+    "@manuscripts/data": "1.0.1",
```
mnatsheh, line 33: I believe that there is a 1.0.2 version of data, so let's use that. Also, you should have a caret there.

mnatsheh (changes requested):

package.json
mnatsheh: Actually, we already have https://github.com/Atypon-OpenSource/manuscripts-transform/pull/18 which updates manuscripts-data to 1.0.2. We should just wait for that.

PR #22, opened 2023-05-07
LEAN-2357

mbetamony:
for this ticket, i did the following: 
1- wrap sections into the 3 new section containers when moving sections to body
2- create a `section_container` node to handle nested sections behaviors
Here are some test results I did using `pressroom` (`importJats`/`exportJats`) found in testResults.zip

1. mnTest.zip is the input i used for `importJats`
2. manuscript-2.manuproj is the output of `importJats`, also the input of `exportJats`. Please note the `abstracts`, `body` and `backmatter` sections with their `elementIDs`
3. exportJatsOutput.xml is the output of `exportJats`. Note how the sections are surrounded between the new sections tags
(search for 
`abstracts` or `body` or `backmatter`)

[testResults.zip](https://github.com/Atypon-OpenSource/manuscripts-transform/files/11415609/testResults.zip)

Please let me know if we want to move on with this implementation or if you require any changes, so I can proceed with adding the necessary tests

Note: 
I realize the branch name is wrong, as I had some issue with GitHub. I'll fix that ASAP 
(LEAN is spelled LEA)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@manuscripts/transform",
   "description": "ProseMirror transformer for Manuscripts applications",
-  "version": "1.1.6",
+  "version": "1.1.7-LEAN-2357",
   "repository": "github:Atypon-OpenSource/manuscripts-transform",
   "license": "Apache-2.0",
   "main": "dist/cjs",
--- a/src/jats/importer/jats-body-transformations.ts
+++ b/src/jats/importer/jats-body-transformations.ts
@@ -24,6 +24,19 @@ const removeNodeFromParent = (node: Element) =>
 const capitalizeFirstLetter = (str: string) =>
   str.charAt(0).toUpperCase() + str.slice(1)
 
+const createSectionContainer = (
+  type: string,
+  createElement: (tagName: string) => HTMLElement
+) => {
+  const sectionContainer = createElement('sec')
+  const category = chooseSectionCategoryByType(type)
+  sectionContainer.setAttribute(
+    'sec-type',
+    category ? chooseSecType(category) : ''
+  )
+  return sectionContainer
+}
+
 export const jatsBodyTransformations = {
   ensureSection(
     body: Element,
@@ -194,19 +207,34 @@ export const jatsBodyTransformations = {
     references: BibliographyItem[] | null,
     createElement: (tagName: string) => HTMLElement
   ) {
+    const bodyContainer = createSectionContainer('body', createElement)
+    const abstractsContainer = createSectionContainer(
+      'abstracts',
+      createElement
+    )
+    const backmatterContainer = createSectionContainer(
+      'backmatter',
+      createElement
+    )
+
+    const bodySections = doc.querySelectorAll('body > sec')
+    for (const section of bodySections) {
+      removeNodeFromParent(section)
+      bodyContainer.append(section)
+    }
     const abstractNodes = doc.querySelectorAll(
       'front > article-meta > abstract'
     )
     for (const abstractNode of abstractNodes) {
       const abstract = this.createAbstract(abstractNode, createElement)
       removeNodeFromParent(abstractNode)
-      body.insertBefore(abstract, body.firstChild)
+      abstractsContainer.appendChild(abstract)
     }
 
     // move sections from back to body
     for (const section of doc.querySelectorAll('back > sec')) {
       removeNodeFromParent(section)
-      body.appendChild(section)
+      backmatterContainer.appendChild(section)
     }
 
     // move acknowledg(e)ments from back to body section
@@ -217,7 +245,7 @@ export const jatsBodyTransformations = {
         createElement
       )
       removeNodeFromParent(ackNode)
-      body.appendChild(acknowledgements)
+      backmatterContainer.appendChild(acknowledgements)
     }
 
     //move appendices from back to body
@@ -226,12 +254,17 @@ export const jatsBodyTransformations = {
     for (const app of appGroup) {
       const appendix = this.createAppendixSection(app, createElement)
       removeNodeFromParent(app)
-      body.appendChild(appendix)
+      backmatterContainer.appendChild(appendix)
     }
     // move bibliography from back to body section
     if (references) {
-      body.appendChild(this.createBibliography(doc, references, createElement))
+      backmatterContainer.appendChild(
+        this.createBibliography(doc, references, createElement)
+      )
     }
+    body.insertBefore(abstractsContainer, body.firstChild)
+    body.appendChild(bodyContainer)
+    body.appendChild(backmatterContainer)
   },
   mapFootnotesToSections(
     doc: Document,
--- a/src/jats/jats-exporter.ts
+++ b/src/jats/jats-exporter.ts
@@ -912,6 +912,7 @@ export class JATSExporter {
       id ? (this.modelMap.get(id) as T | undefined) : undefined
 
     const nodes: NodeSpecs = {
+      section_container: (node) => ['sec', { group: node.attrs.group }, 0],
       attribution: () => ['attrib', 0],
       bibliography_element: () => '',
       bibliography_item: () => '',
--- a/src/transformer/encode.ts
+++ b/src/transformer/encode.ts
@@ -253,6 +253,17 @@ const childElements = (node: ManuscriptNode): ManuscriptNode[] => {
 
   return nodes
 }
+const sectionContainerChildElements = (
+  node: ManuscriptNode
+): ManuscriptNode[] => {
+  const nodes: ManuscriptNode[] = []
+  node.forEach((childNode) => {
+    if (isSectionNode(childNode)) {
+      nodes.push(childNode)
+    }
+  })
+  return nodes
+}
 
 const attributeOfNodeType = (
   node: ManuscriptNode,
@@ -662,6 +673,14 @@ const encoders: NodeEncoderMap = {
     paragraphStyle: node.attrs.paragraphStyle || undefined,
     placeholderInnerHTML: node.attrs.placeholder || '',
     quoteType: 'pull',
+  }),
+  section_container: (node, parent, path, priority): Partial<Section> => ({
+    category: buildSectionCategory(node),
+    priority: priority.value++,
+    path: path.concat([node.attrs.id]),
+    elementIDs: sectionContainerChildElements(node)
+      .map((childNode) => childNode.attrs.id)
+      .filter((id) => id),
   }),
   section: (node, parent, path, priority): Partial<Section> => ({
     category: buildSectionCategory(node),
--- a/src/transformer/section-category.ts
+++ b/src/transformer/section-category.ts
@@ -57,6 +57,9 @@ export type SectionCategory =
   | 'MPSectionCategory:supplementary-material'
   | 'MPSectionCategory:supported-by'
   | 'MPSectionCategory:ethics-statement'
+  | 'MPSectionCategory:body'
+  | 'MPSectionCategory:abstracts'
+  | 'MPSectionCategory:backmatter'
 
 export type SecType =
   | 'abstract'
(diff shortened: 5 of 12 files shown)


src/jats/importer/jats-body-transformations.ts
```diff
@@ -194,19 +207,34 @@ export const jatsBodyTransformations = {
...
+    const backmatterContainer = createSectionContainer(
+      'backmatter',
+      createElement
+    )
+
+    const bodySections = doc.querySelectorAll('body > sec')
+    for (const section of bodySections) {
+      removeNodeFromParent(section)
+      bodyContainer.append(section)
+    }
```
asouqi, line 224: I would prefer to separate this logic of wrapping all sections with our own special section to a new method. and we will call it after `moveSectionsToBody`

asouqi (changes requested):
@mbetamony could you please run `yarn test -u` to update snapshots, and you will need to update this [sample](https://github.com/Atypon-OpenSource/manuscripts-transform/blob/86bd2b5e77d0a6bdd36d640b89531799d4bab1b9/src/jats/__tests__/jats-roundtrip.test.ts#L166) for round-trips test case

mbetamony:
@mnatsheh @asouqi 
I published the following changes to the PR: 

1. removed `section_container` node and used the existing `section` node
2. fixed `exportJats` 
4. `section` node now  accepts `bibliography_section` inside its content ( ref-list wasn't showing when  exporting **JATS**)

Please let me know if you have any questions, I'll fix the tests once this approach is agreed on
Thank you

asouqi:
> section node now accepts bibliography_section inside its content ( ref-list wasn't showing when exporting JATS)

also we need to add `footnotes_section`   `graphical_abstract_section`

> I'll fix the tests once this approach is agreed on

for me, I'm ok with that

mbetamony:
I fixed the graphical-abstract section
for the footnotes, we're adding  them to the body section before  `moveSectionsToBody` in the `jats-body-transformations.ts`, so I fixed that. I didn't have to modify sections node for this
I'll test on more files and see if we need to change anything

mbetamony:
refactored some of the code, I didn't like the way I fixed fn sections before. 
I think the current code should work with everything, please let me know if you have any files/tests you'd like me to run 
I'll also fix the current test files, and fix merge conflicts

mbetamony:
There is a bug in graphical abstract sections, i'll  fix it ASAP.

mbetamony:
- Fixed all bugs I was able to find through testing
- Updated snapshots

src/schema/nodes/section.ts
```diff
@@ -58,7 +58,8 @@ const choosePageBreakStyle = (element: HTMLElement): number => {
 
 export const section: NodeSpec = {
   // NOTE: the schema needs paragraphs to be the default type, so they must explicitly come first
-  content: 'section_label? section_title (paragraph | element)* section*',
+  content:
+    'section_label? section_title (paragraph | element)* (graphical_abstract_section | section | footnotes_section | bibliography_section)*',
```
mnatsheh, line 62: Instead of listing the different section types, can we use a group instead? if all these nodes have a group "sections", then we should be able to use "sections*".

src/transformer/section-category.ts
```diff
@@ -168,7 +174,7 @@ export const buildSectionCategory = (
       return 'MPSectionCategory:abstract-graphical'
 
     default:
-      return node.attrs.category || undefined
+      return node.attrs.category || node.attrs.group || undefined
```
mnatsheh, line 177: Why is this needed?

src/jats/importer/jats-body-transformations.ts
```diff
@@ -24,6 +24,23 @@ const removeNodeFromParent = (node: Element) =>
...
+) => {
+  const sectionContainer = createElement('sec')
+  const category = chooseSectionCategoryByType(type)
+  sectionContainer.setAttribute(
+    'sec-type',
+    category ? chooseSecType(category) : ''
+  )
+
+  const title = createElement('title')
+  title.textContent = category ? chooseSecType(category) : ''
```
mnatsheh, line 39: Can we use the section category title here?

src/jats/importer/jats-body-transformations.ts
```diff
@@ -194,19 +211,40 @@ export const jatsBodyTransformations = {
     references: BibliographyItem[] | null,
     createElement: (tagName: string) => HTMLElement
   ) {
+    const bodyContainer = createSectionContainer('body', createElement)
+    const abstractsContainer = createSectionContainer(
+      'abstracts',
+      createElement
+    )
+    const bodySections = doc.querySelectorAll(
+      'body > sec:not([sec-type="backmatter"]), body > sec:not([sec-type])'
```
mnatsheh, line 220: Is `sec[sec-type="backmatter"]` something that actually exist in the content?

src/jats/jats-exporter.ts
```diff
@@ -2025,29 +2024,52 @@ export class JATSExporter {
       table.insertBefore(tfoot, tbody as Element)
     }
   }
-
+  private moveBody = (body: HTMLElement) => {
```
mnatsheh, line 2027: Let's call this `unwrapBody`.

src/jats/jats-exporter.ts
```diff
@@ -2025,29 +2024,52 @@ export class JATSExporter {
...
+    const container = body.querySelector(':scope > sec[sec-type="body"]')
+    if (!container) {
+      return
+    }
+    const sections = body.querySelectorAll(
```
mnatsheh, line 2032: You should be able to query `container` directly: `container.querySelectorAll('> sec')`.

src/jats/jats-exporter.ts
```diff
@@ -2025,29 +2024,52 @@ export class JATSExporter {
...
+      ':scope > sec[sec-type="body"] > sec'
+    )
+    sections.forEach((section) => {
+      body.appendChild(section.cloneNode(true))
+    })
+    body.removeChild(container)
+  }
+  private removeBackContainer = (body: HTMLElement) => {
+    const container = body.querySelector(':scope > sec[sec-type="backmatter"]')
```
mnatsheh, line 2041: So we're assuming everything has been moved out of this section. Let's check if it is actually empty and log a warning if not.

src/jats/jats-exporter.ts
```diff
@@ -2025,29 +2024,52 @@ export class JATSExporter {
...
-        sectionType === 'abstract-graphical'
-      ) {
-        return true
-      }
+    const container = body.querySelector(':scope > sec[sec-type="abstracts"]')
+    const sections = container
+      ? body.querySelectorAll(':scope > sec[sec-type="abstracts"] > sec')
+      : body.querySelectorAll(':scope > sec')
+
+    const abstractSections = container
```
mnatsheh, line 2052: I prefer we combine those into a single if statement:

```
let abstractSections
if (container) {
    abstractSections = ...
} else {
    abstractSections = ...
}
```

src/transformer/encode.ts
```diff
@@ -671,7 +692,7 @@ const encoders: NodeEncoderMap = {
       inlineContentsOfNodeType(node, node.type.schema.nodes.section_label) ||
       undefined,
     path: path.concat([node.attrs.id]),
-    elementIDs: childElements(node)
+    elementIDs: sectionChildElements(node)
       .map((childNode) => childNode.attrs.id)
       .filter((id) => id),
```
mnatsheh, line 697: Can we move these into the method? We can rename it to `sectionChildElementIds`.

src/transformer/encode.ts
```diff
@@ -253,6 +256,24 @@ const childElements = (node: ManuscriptNode): ManuscriptNode[] => {
...
+const sectionChildElements = (node: ManuscriptNode): ManuscriptNode[] => {
+  const nodes: ManuscriptNode[] = []
+
+  node.forEach((childNode) => {
+    if (
+      !(
+        isSectionNode(childNode) ||
+        isBibliographySectionNode(childNode) ||
+        isGraphicalAbstractSectionNode(childNode) ||
+        isFootnotesSectionNode(childNode)
```
mnatsheh, line 268: Can we check the group of the section category instead of checking the types one by one?

mnatsheh (changes requested):

src/jats/importer/jats-body-transformations.ts
mbetamony: A couple of questions on this:
Do you want me to bring them from manuscripts-data or or use a switch statement similar to [chooseSectionCategoryFromTitle](https://github.com/Atypon-OpenSource/manuscripts-transform/blob/1556eb5b3613409da630626c3ed12de60c54ef0f/src/transformer/section-category.ts#L306)  or something else?
In manuscripts-data, the titles have been left empty for the new sections. I think its better If I add the following titles:
- backmatter -> backmatter
- body -> body
- abstracts-> abstracts

I'm not sure about capitalizing the first letter of the title, there are some inconsistencies in the data
Should I proceed with this change ?

src/schema/nodes/section.ts
mbetamony: Done

src/transformer/encode.ts
mbetamony: Done

src/transformer/section-category.ts
mbetamony: This was left unintentionally when removing the section_container node.

src/jats/importer/jats-body-transformations.ts
mbetamony: Yes. 
Before `moveSectionsToBody` we're calling `mapFootnotesToSections`. Which appends some of the `backmatter` sections to the body. So I created the `backmatter` container there so I don't have to query the sections again when moving sections to body.

asouqi:
> Updated snapshots, but what updates do you mean? moving sections to their correct place?

no need as all test pass

src/jats/importer/jats-body-transformations.ts
mnatsheh: You should use whatever is in manuscripts-data. It's fine if it's empty for now (I'm assuming that it will not break with an empty title)

src/jats/jats-exporter.ts
mnatsheh: Yeah that's fine.

src/jats/jats-exporter.ts
```diff
@@ -2025,30 +2024,57 @@ export class JATSExporter {
...
       }
+      body.removeChild(container)
+    }
+  }
+  private moveAbstracts = (front: HTMLElement, body: HTMLElement) => {
+    const container = body.querySelector(':scope > sec[sec-type="abstracts"]')
+    let abstractSections
+    if (container) {
+      abstractSections = Array.from(
+        body.querySelectorAll(':scope > sec[sec-type="abstracts"] > sec')
```
mnatsheh, line 2053: Let's do this the same way we did it for body (query the container).

src/jats/importer/jats-body-transformations.ts
mnatsheh: Can we call `mapFootnotesToSections` inside `moveSectionsToBody`? I don't see a reason why it's separate. This was we can create all three containers in the same place.

mnatsheh (changes requested):

src/jats/importer/jats-body-transformations.ts
mbetamony: Sure, I also did refactored `moveSectionsToBody` into multiple functions for better readability, let me know if you prefer the old code.

mbetamony:
Sorry for the repeated commits, not sure what happened when I pushed this morning 
Do you wish I get rid of them?

package.json
```diff
@@ -29,7 +29,7 @@
     "version": "yarn build"
   },
   "dependencies": {
-    "@manuscripts/json-schema": "^2.0.0",
+    "@manuscripts/json-schema": "^2.0.1",
```
asouqi, line 32: use `2.0.1-LEAN-2416`

asouqi (approved):

package.json
mbetamony: Done

hhusban (approved):

mnatsheh (approved):

PR #23, opened 2023-05-08
LEAN-2495 - update json-schema version

mnatsheh:


mnatsheh:
The pipeline is failing because of a Jenkins issue, but it made sure things are working locally, so I'll proceed with the merge.

PR #27, opened 2023-07-09
LEAN-2615

mbetamony:
Removed leading & trailing white-space when retrieving attributes. 
I used `getTrimmedAttributeNS` and `getTrimmedAttribute`  instead of directly calling `.trim()` so we ensure it's applied consistently everywhere the function is used, however if you prefer we use `.trim()` directly I can do that.


asouqi (approved):

mbetamony:
Based on my discussion with @hhusban I've been actually removing spaces from attributes and not actual elements 
I'll proceed with removing leading/trailing spaces from elements but should I keep this change @mnatsheh

mbetamony:
Latest commit reverts all changes from before, and trims text whenever we extract textContent from elements.

mnatsheh (approved):

PR #31, opened 2023-07-26
LIT-528106 Allow handling of duplicated ids

mbartenev-atypon:


hhusban (approved):
Looks good

mnatsheh:
@mbartenev-atypon is this the only PR related to the duplicate IDs problem?

mbartenev-atypon:
> @mbartenev-atypon is this the only PR related to the duplicate IDs problem?

Here is the other one
https://github.com/Atypon-OpenSource/manuscripts-article-editor/pull/239/files

mbartenev-atypon:
@mnatsheh looks like we don't need that PR anymore. Closing.

mnatsheh:
@mbartenev-atypon was the fix included somewhere else? did we handle the issue in some other way?

mbartenev-atypon:
hm, I was under impression that it was used on some other branch for the purposes of testing zax and was merged then. But now I'm checking the code and I see that's not on master so it was never merged, although the ticket is closed. Let me know if we need to merged it.

mnatsheh:
I can still see tickets being filed with the "duplicate ids in model map" error (e.g. LEAN-3047).

PR #32, opened 2023-07-27
LEAN-2647

asouqi:


src/schema/nodes/bibliography_item.ts
```diff
@@ -38,6 +38,7 @@ export interface BibliographyItemNode extends ManuscriptNode {
 }
 
 export const bibliographyItem: NodeSpec = {
+  content: 'inline{0}',
```
mbartenev-atypon, line 41: why do we need to specify that the node has empty inline content? will it not be empty without content definition by default?

asouqi: The decoration widget will not be able to reach the bibliography item without content, I know it's the same as removing content, but in some way it will help us to reach the HTML of that node so we could add comment icon

mbartenev-atypon: I see, can you please add that as a comment here. Thanks!

asouqi: sure, updated

mbartenev-atypon (approved):

PR #33, opened 2023-07-30
LEAN-2650: manuscripts-data elimination 

mbetamony:


src/lib/utils.ts
```diff
@@ -96,3 +97,8 @@ export const getTrimmedTextContent = (
   }
   return node.querySelector(querySelector)?.textContent?.trim()
 }
+
+export const loadJsonData = (filePath: string) => {
+  const rawData = fs.readFileSync(filePath, 'utf-8')
```
asouqi, line 102: can we avoid using the file system as that will not work with FE, I know that method will be called just from pressroom but the `utils.ts` will be loaded when we use transform in the FE. you can import `core-sections.json` directly

asouqi (changes requested):

src/lib/utils.ts
mbetamony: this would require setting `resolveJsonModule` to `true` in `tsconfig.json` 
is that okay?

asouqi: yes that is ok, @mnatsheh @hhusban do you have any objections about `resolveJsonModule`

asouqi:
@mnatsheh should we configure that value `DEFAULT_BUNDLE` instead of using it from the transform in the [editor](https://github.com/Atypon-OpenSource/manuscripts-body-editor/blob/b737ca1550953564dcc9d376a5a370c654952942/src/views/bibliography_item.tsx#L42)?

mnatsheh:
> @mnatsheh should we configure that value `DEFAULT_BUNDLE` instead of using it from the transform in the [editor](https://github.com/Atypon-OpenSource/manuscripts-body-editor/blob/b737ca1550953564dcc9d376a5a370c654952942/src/views/bibliography_item.tsx#L42)?

Using `DEFAULT_BUNDLE` there is actually a problem (see LEAN-2727). We should be using the bundle from the manuscript.

asouqi (approved):
@mbetamony I'm ok with both options: update `resolveJsonModule` or create `ts` file

mbetamony:
rerun

mnatsheh (approved):
Let's call them core section categories instead of core sections.

mbetamony:
> Let's call them core section categories instead of core sections.

Done, @asouqi I published a new version for the renamed `core sections`
`1.3.6-LEAN-2650-v2`

PR #43, opened 2023-09-14
LIT-537312

mnatsheh:
It looks like when importing JATS that doesn't have references, the resulting manuproj still contains an empty BibliographyItem, which results in invalid JATS on export. 

The other possible fix here is to replace "+" with "*", and maybe even remove "placeholder" all together. @hhusban do you recall what purpose it served initially?


mnatsheh:
It looks like placeholder represents some kind of error, so I just removed it from the model.

PR #50, opened 2023-11-08
LEAN-3014: Move contributor_list & affiliation_list to the body of the document

mbetamony:

PR #52, opened 2023-11-13
LEAN-3030

hhusban:
Update transformer to decode prosemirror models under 3 core sections


src/transformer/section-category.ts
```diff
@@ -220,6 +220,33 @@ export const chooseSecType = (sectionCategory: SectionCategory): SecType => {
   }
 }
 
+export const chooseCoreSectionBySection = (section: string): string => {
+  switch (section) {
```
hhusban, line 224: since the transformer doesn't have access to the Section-Category.json from the API, this made sense, @mnatsheh what do you think?

hhusban:
@mbartenev-atypon There is a slight issue with this PR, the HTML generated is still showing multiple levels of sections, although I did add "atom" attr to the core-section.ts in the schema, any thoughts how to fix this?

mbartenev-atypon:
> @mbartenev-atypon There is a slight issue with this PR, the HTML generated is still showing multiple levels of sections, although I did add "atom" attr to the core-section.ts in the schema, any thoughts how to fix this?

I don't think we should use "atom" when there is editable content inside. I'm not sure what html do you mean. Do you mean html generated by html exporter or PM? According to you changes it's perfectly valid for core_section have sections inside it and sections can have sections inside them. In any case schema and attributes describe the rules but the document is still built by decoder and adjustments need to be made there to match schema rules. Hope that helps!

hhusban:
> I don't think we should use "atom" when there is editable content inside. I'm not sure what html do you mean. Do you mean html generated by html exporter or PM? According to you changes it's perfectly valid for core_section have sections inside it and sections can have sections inside them. In any case schema and attributes describe the rules but the document is still built by decoder and adjustments need to be made there to match schema rules. Hope that helps!

These section are "fake" section, we don't want to have a representation for it in the jats nor in the HTML
>Do you mean html generated by html exporter 
yes

mnatsheh:
As discussed offline, we don't need to worry about the html exporter for now.

mnatsheh (approved):

src/schema/nodes/manuscript.ts
```diff
@@ -25,7 +25,8 @@ export interface ActualManuscriptNode extends ManuscriptNode {
 }
 
 export const manuscript: NodeSpec = {
-  content: '(section | sections)+ meta_section', // sections is a group of elements that have group: sections
+  content:
+    'keywords_section? contributors_section? affiliations_section? abstract_core_section body_core_section backmatter_core_section meta_section',
```
mbetamony, line 29: I think contributors & affiliations need to be at the top of the document

mnatsheh (approved):

src/schema/nodes/manuscript.ts
```diff
@@ -25,7 +25,8 @@ export interface ActualManuscriptNode extends ManuscriptNode {
 }
 
 export const manuscript: NodeSpec = {
-  content: '(section | sections)+ meta_section', // sections is a group of elements that have group: sections
+  content:
+    'contributors_section? affiliations_section? keywords_section? abstract_core_section body_core_section backmatter_core_section meta_section',
```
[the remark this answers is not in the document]
hhusban: sure thing.

hhusban:
[sample-prosemirror-doc.txt](https://github.com/Atypon-OpenSource/manuscripts-transform/files/13516962/sample-prosemirror-doc.txt)

@mbartenev-atypon please take a look at prosemirror schema after this change

mnatsheh:
Some tweaks were needed for the editor to correctly display, PR created for that. I also noted that old content (with sections still in the manuscripts project) do not display correctly; all content is put under the body.

hhusban:
@mnatsheh I think this could be related to the sections priority, I'll look into this.

src/schema/nodes/affiliations.ts
```diff
@@ -13,51 +13,18 @@
...
-
-export interface AffiliationsNode extends ManuscriptNode {
-  attrs: Attrs
-}
+// This node has no representation in json-schema
+// It exists for the purpose of styling in the UI
 
 export const affiliations: NodeSpec = {
-  content: 'section_title? affiliation*',
+  content: 'affiliation*',
```
mbetamony: I remember this caused an issue on the editor, did you try to upload something to the editor? 
https://github.com/Atypon-OpenSource/manuscripts-transform/pull/50#issuecomment-1810880054

mnatsheh: Yes, it works. I'm guessing the fact it was a section before required it to have a title.

mbetamony: Oh.. yeah. Thanks.

src/jats/__tests__/jats-journal-meta.test.ts
```diff
@@ -36,7 +36,7 @@ describe('JATS journal metadata', () => {
       abbreviatedTitle: 'PubMed Abbreviated Journal Title',
     })
     expect(journal.ISSNs).toHaveLength(1)
-    expect(journal.ISSNs![0]).toStrictEqual({ ISSN: '1234-5678' })
+    expect(journal.ISSNs![0]).toEqual({ ISSN: '1234-5678' })
```
mbartenev-atypon, line 39: Why is that changed to toEqual? Do we expect ISSN to be undefined in some cases?

src/jats/jats-exporter.ts
```diff
@@ -1109,7 +1065,7 @@ export class JATSExporter {
 
         return ['sec', attrs, 0]
       },
-      hard_break: () => ['break'],
+      hard_break: () => '',
```
mbartenev-atypon, line 1068: Not sure how it worked before. Will that still ensure a hard break to be properly exported? Or is it not really translatable into JATS?

mnatsheh: Although they are similar, a hard_break is allowed in places where <break> isn't (e.g. paragraphs), so I opted to remove the mapping for the time being since a missing <break> is not going to change the meaning of the output. I will file a ticket to understand more what hard_break does.

src/jats/__tests__/jats-journal-meta.test.ts
mnatsheh: Changes to the parsing code resulted in the parsed ISSN to have an undefined "publicationType" field. The ISSN itself will not be undefined.

mbartenev-atypon (approved):

PR #53, opened 2023-11-15
LEAN-3034: Exporting citations 

mbetamony:
Once this is approved and the implementation is agreed on, I'll create a ticket for the WM to send the citations style & locale. 
Most of the current implementation is based on what we have in manuscripts-library.


mnatsheh:
I'm wondering if it's time to merge manuscripts-library into manuscripts-transform. I don't see any case where we would want transform without the logic of library. @IslamJMomani @mbartenev-atypon

IslamJMomani:
> I don't see any case where we would want transform without the logic of library

I didn't get that. Could you please elaborate more on the benefits of doing that?

In general, I'm not in favor of merging these modules. I would prefer to keep them separate.

mbartenev-atypon:
> I'm wondering if it's time to merge manuscripts-library into manuscripts-transform. I don't see any case where we would want transform without the logic of library. @IslamJMomani @mbartenev-atypon

I see that the library comes along with transform normally, but unless there is a risk of some convoluted interdependency between these modules, I would prefer to keep them separated. This is because transform is a huge module by itself and library receives it's own specific updates frequently. So mostly for house keeping reasons, so to say, to me it makes more sense to have 2 modules.

mnatsheh:
The main reason I'm suggesting merging the two modules is that now we need "library" logic in the transform, but since library has a dependency on transform, we can't add library as a dependency to transform. The options we have are:
# Duplicate the code. This is what was done in this PR.
# Remove the transform dependency from library. This way we can add library as a dependency to transform and share the code.
# Merge the modules. This is my suggestion. 

Note that until a year ago, there was a cyclic dependency between the modules, which indicates to me that there is a boundary problem with the modules. I think the problem is that transform contains too much, for example, if we extracted the JATS logic into its own module, we wouldn't need this cyclic dependency. I guess that's a 4th option we can pursue.

Thoughts? @IslamJMomani @mbartenev-atypon @hhusban @mbetamony

mbartenev-atypon:
"I would vote for the second option. Removing the transform dependency from the library seems like an achievable task, and sharing the code is the right way to go."
- I agree

mbetamony:
> w issue
> Edit
> Hide
> Delete

mbetamony:
> Any thoughts on moving JATS code out of transform? This way:
> 
> * We can have the transform focused on json schema to prosemirror transformation.
> * No need to add library dependency to transform, since JATS is the only user.

I do agree with separating the import/export logic from the encode/decode logic.

mnatsheh:
I scheduled a quick call tomorrow to discuss this.

mnatsheh:
I thought about it more, and extracting the JATS code to its own module feels like something we should do, as transform just has too much in it. 

If there is consensus by the team around another solution, I'm happy to go with that.

mnatsheh:
I also experimented with removing the transform dependency from library, and although most of it was smooth (most usages should not be there in the first place), there were a couple of usages that were problematic: `buildBibliographicName` and `buildBibliographicDate`. This is another example of logic that (to me) doesn't fit in the transform; it feels more appropriate to have the builders defined in json-schema.

mnatsheh:
@mbetamony I'm assuming the current implementation functions correctly. Maybe having the duplication here is best way forward until we figure out the proper way to fix it.

mbartenev-atypon:
As long as we can preserve testing suites in the transform as we have them now, I'm ok with this.

mbetamony:
Can someone please review this?

mnatsheh:
I've created PRs to add builders to json-schema, and remove transform dependency from library, so it should be possible to use the library here as follows:

```
// setup the provider. Should be done once. See the bibliography plugin in body-editor as an example
const provider = new CitationProvider({
  getLibraryItem: (id: string) => modelMap.get(id),
  citationStyle: csl.style,
  locale: csl.locale,
})

// collect citation nodes in order (using node.descendents)
const nodes = ...
const citations = buildCitations(nodes)
const citationTexts = provider.rebuildState(citations)

----

// "citation" node handler
const id = node.attrs.id
const text = citationTexts.get(id)

----

// when building the ref-list. This ensures only cited references are there, and that they're in correct order.
const [meta] = provider.makeBibliography()
for (const id of meta.entry_ids) {
    ...
}

File: src/jats/jats-exporter.ts [hunk 1]
```diff
@@ -74,13 +76,13 @@ interface Links {
     [key: string]: string
   }
 }
+type CitationNodes = [CitationNode, number][]
```
mnatsheh, line 79: This should not be needed. The number (pos) is not relevant for exporting.

src/jats/jats-exporter.ts
```diff
@@ -925,8 +971,8 @@ export class JATSExporter {
         const xref = this.document.createElement('xref')
         xref.setAttribute('ref-type', 'bibr')
         xref.setAttribute('rid', normalizeID(rids.join(' ')))
-
-        if (citation.attrs.contents) {
+        xref.textContent = this.citationTexts.get(node.attrs.id) ?? ''
+        if (node.attrs.contents) {
           // TODO: convert markup to JATS?
           // xref.innerHTML = node.attrs.contents
           const text = textFromHTML(node.attrs.contents)
```
mnatsheh, line 978: This whole if statement should go away.

mnatsheh (changes requested):
As I mention before, we should also update the <ref-list> generation to use the citation provider.

src/jats/jats-exporter.ts
mbetamony: Yes I'm still not done with this PR.

mbetamony: I removed it, I'm also throwing an error if no match is found in the text map.

src/jats/jats-exporter.ts
```diff
@@ -716,18 +756,14 @@ export class JATSExporter {
...
     back.appendChild(refList)
-
-    const bibliographyItems = this.models.filter(
-      hasObjectType<BibliographyItem>(ObjectTypes.BibliographyItem)
-    )
-
-    for (const bibliographyItem of bibliographyItems) {
+    const [meta] = this.citationProvider.makeBibliography()
+    for (const id of meta.entry_ids) {
+      const bibliographyItem = this.modelMap.get(id[0]) as BibliographyItem
```
mnatsheh, line 761: Is `id` an array? In the body editor it is treated as a `string`.

mbetamony: yes, 
image

mnatsheh: That is... weird.

mnatsheh (approved):

PR #54, opened 2023-11-15
LEAN-3014: add affiliations_list & contributors_list back

mbetamony:
@hhusban  please note that I added these back because old documents with affiliation_list & contributor_list won't work.


mnatsheh (approved):
Let's not merge this unless it's absolutely necessary. If LEAN-3030 gets merged, I don't think this will be needed.

mbetamony:
> Let's not merge this unless it's absolutely necessary. If LEAN-3030 gets merged, I don't think this will be needed.

If we're clearing all old submissions after Hashem's update, then yes we don't need to merge this.

mbetamony:
@mnatsheh just a few things i noticed when i tried Zax: 
1- It publishes the branch, then merges it into master. I'm not sure if that's done on purpose or not. Should we publish master after its done merging?
2- I did a minor release for this branch, and all other repos that have transformer as a dependency were published with a minor release as well. Shouldn't it be just a patch for other repos?

mnatsheh:
No longer needed.

PR #55, opened 2023-11-21
LEAN-3092 -- Equations clean-up - BE implementation

hhusban:


src/transformer/encode.ts
```diff
@@ -552,33 +543,21 @@ const encoders: NodeEncoderMap = {
         : false,
   }),
   equation: (node): Partial<Equation> => ({
-    MathMLStringRepresentation:
-      node.attrs.MathMLStringRepresentation || undefined,
-    TeXRepresentation: node.attrs.TeXRepresentation,
-    SVGStringRepresentation: node.attrs.SVGStringRepresentation,
-    // title: 'Equation',
+    content: node.attrs.content,
```
asouqi, line 546: we need a json-schema update to remove `MathMLStringRepresentation`, `TeXRepresentation`, `SVGStringRepresentation` and replace them with content

asouqi (dismissed):

src/transformer/encode.ts
```diff
@@ -552,33 +543,21 @@ const encoders: NodeEncoderMap = {
...
     elementType: 'p',
-    suppressCaption: Boolean(node.attrs.suppressCaption) || undefined,
     suppressTitle:
       node.attrs.suppressTitle === undefined ||
       node.attrs.suppressTitle === true
         ? undefined
         : false,
   }),
+  inline_equation: (node): Partial<InlineMathFragment> => ({
+    content: node.attrs.content,
```
asouqi, line 559: this also needs to be updated from json-schema

src/schema/nodes/equation_element.ts
```diff
@@ -43,7 +43,7 @@ export const equationElement: NodeSpec = {
   group: 'block element',
   parseDOM: [
     {
-      tag: 'figure.equation',
+      tag: 'disp-formula',
```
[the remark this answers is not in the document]
hhusban: This is just a placeholder, this PR still has few things un-finished

src/transformer/encode.ts
hhusban: I already have it locally, but there still other things I'm waiting a confirmation which might also affect the json-schema

hhusban:
> @hhusban MathML will be the only source of truth to delay with equation? and we are going to drop support for svg & latex
Yes, other conversions might still happen, but there is no need to store them.

src/schema/nodes/equation_element.ts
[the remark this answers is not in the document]
mnatsheh: Let's use `div.equation` here.

src/schema/nodes/inline_equation.ts
```diff
@@ -18,27 +18,32 @@ import { ObjectTypes } from '@manuscripts/json-schema'
...
 import { ManuscriptNode } from '../types'
+import { CommentNode } from './comment'
 
 interface Attrs {
   id: string
-  MathMLRepresentation: string
-  SVGRepresentation: string
-  TeXRepresentation: string
+  suppressCaption: boolean
+  suppressTitle?: boolean
```
mnatsheh, line 26: Do we actually need those?

src/schema/nodes/inline_equation.ts
```diff
@@ -18,27 +18,32 @@ import { ObjectTypes } from '@manuscripts/json-schema'
...
+  title: string
```
mnatsheh, line 27: I think this is supposed to be under equation_element.

src/schema/nodes/inline_equation.ts
```diff
@@ -47,37 +52,21 @@ export const inlineEquation: NodeSpec = {
     {
       tag: `span.${ObjectTypes.InlineMathFragment}`,
```
mnatsheh, line 53: Can we change this to `span.math`?

src/schema/nodes/equation.ts
```diff
@@ -17,64 +17,63 @@
...
 
+import { convertTeXToMathML } from '../../mathjax/tex-to-mathml'
 import { ManuscriptNode } from '../types'
 
 interface Attrs {
   id: string
-  MathMLStringRepresentation: string
-  SVGStringRepresentation: string
-  TeXRepresentation: string
+  content: string
```
mnatsheh, line 25: Let's use "contents" here for consistency with other models.

src/schema/nodes/inline_equation.ts
```diff
@@ -18,27 +18,32 @@ import { ObjectTypes } from '@manuscripts/json-schema'
...
 
 interface Attrs {
   id: string
-  MathMLRepresentation: string
-  SVGRepresentation: string
-  TeXRepresentation: string
+  suppressCaption: boolean
+  suppressTitle?: boolean
+  title: string
+  content: string
```
mnatsheh, line 28: Same as equation.

src/schema/nodes/inline_equation.ts
```diff
@@ -18,27 +18,32 @@ import { ObjectTypes } from '@manuscripts/json-schema'
 import { NodeSpec } from 'prosemirror-model'
 
 import { ManuscriptNode } from '../types'
+import { CommentNode } from './comment'
 
 interface Attrs {
   id: string
```
mnatsheh, line 24: We don't need an id for inline math.

src/transformer/decode.ts
```diff
@@ -348,19 +348,25 @@ export class Decoder {
...
       ) as EquationElementNode
     },
+    [ObjectTypes.InlineMathFragment]: (data) => {
+      const model = data as InlineMathFragment
+
+      return schema.nodes.inline_equation.createChecked({
+        id: model._id,
+        content: model.content,
+      }) as InlineEquationNode
+    },
```
mnatsheh, line 369: This should not be needed; Instead, we should skip encoding inline math in its own model. In encode, I added a check to skip all children of paragraph except inline math, so we should update it. Maybe we can try replacing that check with another that skips all inline nodes, since inline by definition means it should not be a standalone model.

src/transformer/encode.ts
```diff
@@ -552,33 +543,21 @@ const encoders: NodeEncoderMap = {
...
+  }),
```
mnatsheh, line 560: See comment in decode.

src/schema/nodes/equation.ts
```diff
@@ -17,64 +17,63 @@
 import { ObjectTypes } from '@manuscripts/json-schema'
 import { NodeSpec } from 'prosemirror-model'
 
+import { convertTeXToMathML } from '../../mathjax/tex-to-mathml'
 import { ManuscriptNode } from '../types'
 
 interface Attrs {
```
mnatsheh, line 23: Let's add a "format" attribute to both this and inline_equation (and json-schema). The allowed values should be "tex" and "mathml".

src/schema/nodes/equation.ts
```diff
@@ -17,64 +17,63 @@
...
+  const container = element.querySelector('alternatives') ?? element
+  let content: string | null = ''
+  for (const child of container.childNodes) {
+    // remove namespace prefix
+    // TODO: real namespaces
+    const nodeName = child.nodeName.replace(/^[a-z]:/, '')
+
+    switch (nodeName) {
+      case 'tex-math':
+        content = convertTeXToMathML((child as Element).outerHTML, true)
```
mnatsheh, line 43: Thinking about it more, I feel that we shouldn't do this conversion. In transform we should keep the data as is, and the FE can handle the different cases.

mnatsheh (changes requested):

src/schema/nodes/inline_equation.ts
mnatsheh: What do you mean by "citations"?

hhusban: nevermind, my test was flawed, i removed the id

mnatsheh:
Are there still usages for mathjax? if not, can we remove the dependency?

src/jats/__tests__/jats-exporter.test.ts
```diff
@@ -800,9 +800,8 @@ describe('JATS exporter', () => {
     projectBundle.data.find((el) => {
       if (el._id === 'MPEquation:C900DDA4-BE45-4AF6-8C9F-CA0AA5FCC403') {
         const equation = el as Equation
-        equation.MathMLStringRepresentation = mathMLFragment
+        equation.content = mathMLFragment
         // @ts-ignore
-        delete equation.TeXRepresentation
       }
```
mnatsheh, line 805: I'm guessing this if statement can be removed; it's not updating the correct field.

src/jats/importer/jats-body-dom-parser.ts
```diff
@@ -41,6 +39,27 @@ const chooseContentType = (graphicNode?: Element): string | undefined => {
...
+  for (const child of container.childNodes) {
+    // remove namespace prefix
+    // TODO: real namespaces
+    const nodeName = child.nodeName.replace(/^[a-z]:/, '')
+
+    switch (nodeName) {
+      case 'tex-math':
+      case 'mml:math':
+        contents = (child as Element).outerHTML
+        format = nodeName
```
mnatsheh, line 56: Let's have the format values be "mathml" and "tex".

src/schema/nodes/equation.ts
```diff
@@ -44,37 +42,27 @@ export const equation: NodeSpec = {
...
       },
     },
-    // TODO: convert MathML from pasted math elements?
   ],
-  toDOM: (node) => {
+  toDOM: (node: ManuscriptNode) => {
     const equationNode = node as EquationNode
+    const { id, contents, format } = equationNode.attrs
 
+    const attrs: { [key: string]: string } = {}
```
mnatsheh, line 58: It doesn't seem `attrs` is being used.

src/schema/nodes/equation_element.ts
```diff
@@ -31,19 +30,18 @@ export interface EquationElementNode extends ManuscriptNode {
...
+    label: { default: '' },
     dataTracked: { default: null },
     comments: { default: null },
   },
   selectable: false,
   group: 'block element',
   parseDOM: [
     {
-      tag: 'figure.equation',
+      tag: 'dev.equation',
```
mnatsheh, line 44: Typo.

src/schema/nodes/inline_equation.ts
```diff
@@ -14,70 +14,56 @@
...
         }
       },
     },
-    // TODO: convert MathML from pasted math elements?
   ],
   toDOM: (node) => {
     const inlineEquationNode = node as InlineEquationNode
+    const { contents, format } = inlineEquationNode.attrs
 
+    const attrs: { [key: string]: string } = {}
```
mnatsheh, line 60: It doesn't seem `attrs` is being used.

mnatsheh (approved):

PR #58, opened 2023-11-30
LEAN-3122: add placeholder_element to affiliatons_section & contributors_section

mbetamony:
I think the error is happening because we don't have placeholder element inside the content of these sections.
https://lwf-ceres-dev5.literatumonline.com/editor/a02eb931-6412-4b97-bf7b-dda7ef6919b1

Screenshot from 2023-11-30 17-37-52

image


mnatsheh:
Where is the placeholder_element coming from? I would rather we not generate it in the first place.

mbetamony:
> Where is the placeholder_element coming from? I would rather we not generate it in the first place.

I'm not really how sure how the frontend handles the deletion of contributor node, but i'll try to find out.
https://jira.atypon.com/secure/attachment/1647474/1647474_submission+is+not+loading+after+deleting+author.webm

mnatsheh:
Shouldn't the changes from LEAN-3030 fix this? after that we should no longer have json schema models for these sections.

mbetamony:
> Shouldn't the changes from LEAN-3030 fix this? after that we should no longer have json schema models for these sections.

Yes that's correct, i'll keep the PR open until it is retested.

mnatsheh:
No longer needed.

PR #64, opened 2024-01-02
Adjust TableElement content order

IslamJMomani:
Based on the designs, the table element footer should be displayed directly after the table and before the caption.

Screenshot 2024-01-11 at 11 59 42 AM


mbetamony (approved):

IslamJMomani:
@mbetamony, thank you for the code review. Do the snapshot changes make sense to you? I can't explain them.

mbetamony:
Good catch, I didn't actually look at the tests before (apologies) 
Removing this would probably fix it https://github.com/Atypon-OpenSource/manuscripts-transform/blob/51b7a7be64373cae3abe82bd2eb1a533f10aac6c/src/jats/importer/parse-jats-article.ts#L125 

I'll add Hashem as a reviewer as well for confirmation

src/schema/nodes/table_element.ts
```diff
@@ -38,7 +38,7 @@ export interface TableElementNode extends ManuscriptNode {
 
 export const tableElement: NodeSpec = {
   content:
-    '(table | placeholder) figcaption? table_element_footer? (listing | placeholder)',
+    '(table | placeholder) table_element_footer? figcaption? (listing | placeholder)',
```
hhusban, line 41: This order should also be reflected to the dom being passed to the body-parser

IslamJMomani:
> Good catch, I didn't actually look at the tests before (apologies) Removing this would probably fix it
> 
> https://github.com/Atypon-OpenSource/manuscripts-transform/blob/51b7a7be64373cae3abe82bd2eb1a533f10aac6c/src/jats/importer/parse-jats-article.ts#L125
> 
> I'll add Hashem as a reviewer as well for confirmation

You were right, removing it fixed that problem. Thank you both, @hhusban and @mbetamony, for your guidance.

hhusban (approved):

mbetamony (approved):

mbetamony:
> > Good catch, I didn't actually look at the tests before (apologies) Removing this would probably fix it
> > https://github.com/Atypon-OpenSource/manuscripts-transform/blob/51b7a7be64373cae3abe82bd2eb1a533f10aac6c/src/jats/importer/parse-jats-article.ts#L125
> > 
> > I'll add Hashem as a reviewer as well for confirmation
> 
> You were right, removing it fixed that problem. Thank you both, @hhusban and @mbetamony, for your guidance.

You're most welcome.

PR #65, opened 2024-01-04
LEAN-3030: Fix for circural import causing manuscript to be undefined in dev mode

mbartenev-atypon:


mbartenev-atypon:
@mnatsheh we can merge it without testing, this is a minor fix for a dev env problem only.

PR #66, opened 2024-01-11
LEAN-3083 Authors and Affiliations to Prosemirror

mbartenev-atypon:
Adding some missing properties


mnatsheh (approved):

src/schema/nodes/contributor.ts
```diff
@@ -13,21 +13,23 @@
...
+import { BibliographicName } from '@manuscripts/json-schema'
 import { NodeSpec } from 'prosemirror-model'
 
 import { ManuscriptNode } from '../types'
 
 interface Attrs {
   id: string
   role: string
-  affiliations: Affiliation[]
+  affiliations: string[]
```
asouqi, line 24: are we sure about that type, in the [decode](https://github.com/Atypon-OpenSource/manuscripts-transform/blob/a89e255cd7e3ab0144a0aaadedcc86b148f14965/src/transformer/decode.ts#L840) we pass object array!

asouqi (changes requested):

src/schema/nodes/contributor.ts
mbartenev-atypon: I'm going to check. During runtime I got arrays of ids. So I added these fixes.

mbartenev-atypon: Checked. It's array of string there. 
image

asouqi (approved):

PR #67, opened 2024-01-16
LEAN-3168 Restoring not creating empty section implemented in LEAN-3069

mbartenev-atypon:


mbartenev-atypon:
@mnatsheh should we go with a testing process on this ticket or is it fine to merge straight away?

mnatsheh (approved):
As long as you tried it locally, it should be fine to merge. 

I'm assuming the problem is not actually related to your fix in LEAN-3168.

mbartenev-atypon:
> As long as you tried it locally, it should be fine to merge.
> 
> I'm assuming the problem is not actually related to your fix in LEAN-3168.

Yes, it's not related.

mbartenev-atypon:
I've added that testing to demonstrate that this fix prevents empty section created and so we can monitor that in future.

PR #68, opened 2024-01-17
LEAN-3182 - Add supplements to ProseMirror doc

hhusban:


mnatsheh (approved):
Let's use "supplement" to be consistent with the json-schema. so, we should use `SupplementNode`, `SupplementsNode`, etc.

src/schema/nodes/supplements.ts
```diff
@@ -0,0 +1,56 @@
...
+interface Attrs {
+  id: string
+}
+
+export interface SupplementsNode extends ManuscriptNode {
+  attrs: Attrs
+}
+
+export const supplements: NodeSpec = {
+  content: 'section_title supplement*',
```
asouqi, line 30: @mnatsheh is there going to be a view for supplements in the editor!

mbartenev-atypon:
@mnatsheh @hhusban I picked up that change in the body editor and I see that supplements now are inserted at the beginning of the document right after the keywords. Is that expected?

mnatsheh:
This is not expected. We will include the fix for it in LEAN-3181.

PR #70, opened 2024-02-12
LEAN-3336: TSE-015 - Revisit prosemirror table schema - BE Implementation

mbetamony:
Our schema is similar to  the one applied in [prosemirror-tables](https://github.com/ProseMirror/prosemirror-tables/blob/f438926f02f81bdee08d0725b418f941b1ac29c2/src/schema.ts#L132). However, there are afew differences; our table has a table_colgroup and a table_body instead of directly adding the table rows to the table. And in prosemirror's implementation they use a table_header node for table headers and not regular table_cells like we do.

On the frontend (FE), instead of sending the actual content for the table_body—which aligns with the content of the table in the prosemirror-tables implementation—we send the entire table, which leads to incorrect calculations of width and height for the table.

I tried to match our schema by eliminating table_colgroup and table_body, having table headers, and adding rows directly to the table.  This resolved some problems, like adding a column. But it is still not working correctly and there are a few bugs i think the FE needs to take a look at.

I'm also not sure if we're using colgroup for styling or not. And if we should modify the schema or not, since passing the table_body to prosemirror instead of the entire table would solve the issues fixed here. (we're passing the state)

@mnatsheh,  this is still a work in progress, but please let me know your thoughts and if we need to maintain colgroup or not.


mnatsheh:
I think we should try the following:
- Remove our table nodes, and use `tableNodes` from prosemirror-tables directly.
- Move any custom nodes to `table_element`. We already have the footer there, so let's add the columns there as well.

src/schema/nodes/table.ts
```diff
@@ -14,72 +14,218 @@
  * limitations under the License.
```
mnatsheh, line 14: Is this a copy from the prosemirror-tables?

mbetamony: No, its not. 

I tried to use prosemirror's schema directly, but we use some custom attributes. The first solution was just to copy the schema and add our changes to it (similar to what we had going on before), but maintaining that will become difficult. 

The second solution was to add our custom attributes directly to the objects from the schema, for example replace the toDom() in the table schema to ours. this seemed reasonable but again, we'll have some issues of maintainability. Specially in table_cell & table_header, in toDOM(), we check if the node has textContent `node.textContent` and append  `placeholder` class if it doesn't. And in their implementation and what they expose, we can't check if the node has text content. The only control we have is in the tableOptions we pass to their library to generate the nodes. 

The third solution is what's currently implemented in this PR, it tries to maintain everything their schema has, even if it changes. And adds our own custom changes at the cost of over engineering and a lot of code. 

This is probably easier to explain in a call, so let me know if you have the time. Also please let me know if you prefer something else.

mbetamony:
Can someone review this PR please?
@mnatsheh, @hhusban

PR #74, opened 2024-02-28
TNM-020 - Display uncited notes (LEAN-3377)

asouqi:
one of the [TNM-020](https://jira.atypon.com/browse/LEAN-3377) requirements is to make sure the uncited footnotes in table-footnotes are at the end of the list


hhusban (approved):

asouqi:
@hhusban I changed the process of sorting from `jatsBodyTransformations` to the decode. as when we decode modelMap again there is nothing to tell about order of elements

PR #75, opened 2024-03-03
LEAN-3376: import/export author notes

mbetamony:
Do we need to display the author notes in the editor right now? I believe that with the current styling we have for contributors, they are not showing. (I checked the PM document, and the author notes are there.)


src/jats/importer/parse-jats-article.ts
```diff
@@ -70,6 +71,15 @@ export const parseJATSFront = (doc: Document, front: Element) => {
...
 
+  const authorNotesParagraphs = jatsFrontParser.parseAuthorNotesParagraphs([
+    ...front.querySelectorAll('article-meta > author-notes > p'),
+  ])
+  const authorNotes = buildAuthorNotes(
+    Array.from(footnoteIDs.values()).concat(
+      authorNotesParagraphs.map((p) => p._id)
+    )
+  )
+
```
mnatsheh, line 82: I think we should combine the p and fn parsing in a single `jatsFrontParser.parseAuthorNotes`.

src/jats/jats-exporter.ts
```diff
@@ -1749,57 +1752,29 @@ export class JATSExporter {
...
-            return corresp.correspID
-          })
-          noteIDs.push(...ids)
-        }
-      }
-      const footnotes: Footnote[] = []
-      footnotes.push(
-        ...this.models.filter(hasObjectType<Footnote>(ObjectTypes.Footnote))
+      const authorNotes = this.models.find(
+        hasObjectType<AuthorNotes>(ObjectTypes.AuthorNotes)
```
mnatsheh, line 1756: You can use `getModelsByType` instead.

src/jats/jats-exporter.ts
```diff
@@ -1749,57 +1752,29 @@ export class JATSExporter {
...
+            this.appendParagraphToElement(
+              model as ParagraphElement,
+              authorNotesEl
+            )
+          } else if (id.startsWith('MPCorresponding')) {
+            this.appendCorrespondingToElement(
+              model as Corresponding,
+              authorNotesEl
+            )
+          } else if (id.startsWith('MPFootnote')) {
```
mnatsheh, line 1775: The corresp is not part of the author_notes node, is it?

src/jats/importer/parse-jats-article.ts
mbetamony: A question about this.. are the author-notes optional or required in the jats? we shouldn't generate a model if we don't have an author-notes element right? Or do we generate it anyways?

src/jats/jats-exporter.ts
mbetamony: I kept it in case we do add corresp in the future, but i can remove it.

src/jats/importer/parse-jats-article.ts
mnatsheh: I believe it's optional, and yes we should not generate a model if it doesn't exist.

mnatsheh:
> Do we need to display the author notes in the editor right now? I believe that with the current styling we have for contributors, they are not showing. (I checked the PM document, and the author notes are there.)

yes it will not display in the editor for now.

src/jats/jats-exporter.ts
mbetamony: The reason i used `find` here is to get a single model in the result, and not an array.

src/jats/importer/parse-jats-article.ts
mbetamony: I updated the code, please let me know if any more changes are needed

src/jats/jats-exporter.ts
```diff
@@ -1749,57 +1752,25 @@ export class JATSExporter {
...
-        ...this.models.filter(
-          hasObjectType<Corresponding>(ObjectTypes.Corresponding)
-        )
-      )
-
-      if (footnotes || correspodings) {
+      const authorNotes = getModelsByType<AuthorNotes>(
+        this.modelMap,
+        ObjectTypes.AuthorNotes
+      )[0]
```
mnatsheh, line 1758: Please change this to `)?.[0]` otherwise it will break when no author notes exist.

mnatsheh (approved):

src/jats/jats-exporter.ts
mnatsheh: Oh ok I didn't know that.

mbetamony:
@mnatsheh, please note that I added correps back to the exported jats, some contributors had them referenced and that caused jats validation to fail. https://github.com/Atypon-OpenSource/manuscripts-transform/pull/75/commits/589676abd2810c0b0cd1d14e88d323e9876b3d41
I can also fix it by only adjusting the export jats to what we had previously, but this makes more sense if we're adding corresps to our PM in the future.

src/jats/jats-exporter.ts
```diff
@@ -1879,7 +1858,53 @@ export class JATSExporter {
...
+    if (corresponding.label) {
+      const labelEl = this.document.createElement('label')
+      labelEl.textContent = corresponding.label
+      correspondingEl.appendChild(labelEl)
+    }
+    correspondingEl.append(corresponding.contents)
+    element.appendChild(correspondingEl)
+  }
+
+  private getUsedCorrespondings = (contributors: Contributor[]) => {
```
mnatsheh, line 1876: Can't we do this in a simpler way? Something like:
```
const corresps = []
contributors.forEach((c) => {
   c.corresp.flatMap((cc) => cc.correspID).map((id) => this.modelMap.get(id)).forEach(corresps.push)
})
```

mbetamony: Sure

PR #78, opened 2024-03-24
LEAN-3446: add empty paragraphs to empty footnotes

mbetamony:


mnatsheh:
What makes the JATS invalid in the first place? Can't we just generate an empty `<p>`?

mbetamony:
> What makes the JATS invalid in the first place? Can't we just generate an empty `<p>`?

the empty footnotes, DTD validation result:  `Element fn content does not follow the DTD, expecting (label? , p+)`. Adding an empty paragraph can also fix the issue, i'll do that.

mbetamony:
> What makes the JATS invalid in the first place? Can't we just generate an empty `<p>`?

Updated the code to add empty paragraphs instead of removing the empty footnotes.

mnatsheh (approved):

PR #84, opened 2024-04-14
Import/export corresp

mbetamony:


src/jats/jats-exporter.ts
```diff
@@ -1849,6 +1823,47 @@ export class JATSExporter {
...
+  private appendModelsToAuthorNotes(
+    authorNotesEl: HTMLElement,
+    containedObjectIDs: string[]
+  ) {
+    containedObjectIDs.forEach((id) => {
+      const model = this.modelMap.get(id)
+      if (!model) {
+        return
+      }
+      switch (true) {
```
hhusban, line 1854: Why not use model.objectType?

mbetamony: You're right, that's better.

hhusban (approved):

PR #85, opened 2024-04-15
LEAN-3514

mnatsheh:
As far as I can tell, these are the correct types for these fields. I also don't think the parseDOM and toDOM are necessary here.

### DIFF
--- a/src/schema/nodes/bibliography_item.ts
+++ b/src/schema/nodes/bibliography_item.ts
@@ -17,12 +17,13 @@
 import { NodeSpec } from 'prosemirror-model'
 
 import { ManuscriptNode } from '../types'
+import {BibliographicDate, BibliographicName} from "@manuscripts/json-schema";
 
 interface Attrs {
   id: string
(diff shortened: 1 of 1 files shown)


src/schema/nodes/bibliography_item.ts
```diff
@@ -58,97 +59,4 @@ export const bibliographyItem: NodeSpec = {
   },
   selectable: false,
   group: 'block',
-  parseDOM: [
```
mbartenev-atypon, line 61: Do we not use parse dom for encoding?

mbartenev-atypon (approved):

mbetamony (approved):

hhusban (approved):

src/schema/nodes/bibliography_item.ts
mnatsheh: I don't believe so. I am able to import JATS, edit the content, and export JATS successfully.

PR #86, opened 2024-04-18
Newly Added Text Incorrectly Highlighted as Comment (LEAN-3509)

asouqi:
Adding highlighted text of comment between two **highlightMarker** nodes will lead to adding a normal text between them in the editor or if the user hits enter in that range the new node will have the **highlightMarker** node as described in the issue LEAN-3509. this PR will wrap highlighted text in **highlightMarker** node so we can show it as a static entity like citation,link node.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@manuscripts/transform",
   "description": "ProseMirror transformer for Manuscripts applications",
-  "version": "2.1.12",
+  "version": "2.1.12-LEAN-3509-0",
   "repository": "github:Atypon-OpenSource/manuscripts-transform",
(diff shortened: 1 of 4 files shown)


mnatsheh:
Not sure about this. I don't think we want highlights to be treated like citations, and being able to add text in the middle of highlighted text is reasonable and is how it works in every other editor. The only difference between our editor and others is that when you press Enter _at the end of a paragraph_ the new paragraph will not be highlighted (but if you do it in the middle of a paragraph, the new paragraph will be highlighted).

If it's easy to fix this specific case, let's do it, otherwise let's leave this for now and have a discussion with PdM about it.

PR #88, opened 2024-04-23
LEAN-2746

asouqi:


src/schema/nodes/title.ts
```diff
@@ -30,7 +30,7 @@ export interface TitleNode extends ManuscriptNode {
 }
 
 export const title: NodeSpec = {
-  content: 'text*',
+  content: '(text | highlight_marker)*',
```
mnatsheh, line 33: Can we make this "inline*" instead? Also can we get rid of the explicit marks below?

asouqi: updated

mnatsheh (approved):

asouqi:
@mnatsheh you are right about your concern for the encode/decode, so will need to pass comments when we parse content of title in the decode as we do it for other nodes. and I did update to test case so we could see title comment in both encode/decode

mnatsheh (approved):

PR #89, opened 2024-04-30
LEAN-3519 -- Table footer model rework

hhusban:


mnatsheh (changes requested):

src/schema/nodes/table_element_footer.ts
```diff
@@ -31,7 +31,7 @@ export const tableElementFooter: NodeSpec = {
     id: { default: '' },
     dataTracked: {default: null}
   },
-  content: '(paragraph | footnotes_element)*',
+  content: '(general_table_footnote | footnotes_element)*',
```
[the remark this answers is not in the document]
IslamJMomani: Yes, I believe so

src/jats/importer/jats-body-dom-parser.ts
```diff
@@ -343,6 +343,24 @@ const nodes: NodeRule[] = [
       }
     },
   },
+  {
+    tag: 'p',
+    node: 'general_table_footnote',
+    context: 'footnotes_element/|table_element_footer/',
```
IslamJMomani, line 349: What does that mean? Are you using this 'general_table_footnote' node to represent any **p** in the `table_element_footer` or `footnotes_element`? 

I was thinking that we would only have one 'general_table_footnote' per 'table_element_footer', and it would contain all the paragraphs that do not belong to the 'footnotes_element'

src/jats/__tests__/__snapshots__/jats-importer.test.ts.snap
```diff
@@ -4325,8 +4325,8 @@ Array [
   Object {
     "_id": "MPTableElementFooter:test",
     "containedObjectIDs": Array [
-      "MPParagraphElement:test",
       "MPFootnotesElement:test",
+      "MPParagraphElement:test",
```
IslamJMomani, line 4329: I was expecting an object of type `GeneralTableFootnote` instead.

mnatsheh: Is there a need for a new model? I think we can have the encoder wrap them appropriately.

src/jats/importer/jats-body-dom-parser.ts
nghazzawi-atypon: I believe this modification may impact the `<p>` elements within `footnote_element`, which typically represent references or table footnotes. However, the desired outcome is to exclusively alter the type of all `<p>` elements that are within `table_element_footer` but not  within `footnotes_element`.
so we may have to update the context to be something like : 

context: 'table_element_footer/!footnotes_element'

nghazzawi-atypon (changes requested):

hhusban:
@nghazzawi-atypon @IslamJMomani @mnatsheh I updated the changes to group all paragraph under generalTableFootnote shcema, now waiting for an answer on the inline footnote comment on the ticket.

src/jats/importer/jats-parser-utils.ts
```diff
@@ -56,6 +56,14 @@ const updateNodeID = (
...
     return
   }
+  if (node.type === schema.nodes.general_table_footnote) {
+    // @ts-ignore - while attrs are readonly, it is acceptable to change them when document is inactive and there is no view
+    node.attrs = {
+      ...node.attrs,
+      id: `GeneralTableFootnote:${uuidv4()}`,
+    }
+    return
+  }
```
mnatsheh, line 66: What's the purpose of this?

hhusban: On encoding any node that has no ID will not be iterated on

mnatsheh: can we make the id static like "_"? do we need it to be a uuid?

hhusban: since there might be multiple general_table_footnotes in a single document, not sure if having duplicate IDs would cause any issues in FE side
@IslamJMomani thoughts?

mnatsheh: It's fine, we can leave it as is.

PR #104, opened 2024-06-12
LEAN-3307 SBA-020 - Adding new section in a blank article

fwaisi:


src/schema/nodes/manuscript.ts
```diff
@@ -29,7 +29,7 @@ export interface ActualManuscriptNode extends ManuscriptNode {
 
 export const manuscript: NodeSpec = {
   content:
-    'title? contributors? affiliations? keywords? supplements? abstracts body backmatter comments',
+    'title? contributors? affiliations? keywords? supplements? abstracts? body? backmatter? comments?',
```
mnatsheh, line 32: I don't think there is value in making comments optional. Can we also keep abstracts/body/backmatter required? I also feel title should be required but jats import might break with this.

fwaisi: reverted

PR #105, opened 2024-06-19
LEAN-1330: EI25Q1-030 - Surface non-editable DOI in Editor

mvidtman:


src/schema/nodes/affiliations.ts
```diff
@@ -20,7 +20,7 @@ import { NodeSpec } from 'prosemirror-model'
 // It exists for the purpose of styling in the UI
 
 export const affiliations: NodeSpec = {
-  content: 'affiliation*',
+  content: 'affiliation* paragraph*',
```
mbartenev-atypon, line 23: this change should be only a change in a view of the node, why do we need to change the schema?

src/schema/nodes/manuscript.ts
```diff
@@ -48,11 +50,11 @@ export const manuscript: NodeSpec = {
   ],
   toDOM: (node) => {
     const manuscriptNode = node as ActualManuscriptNode
-
     return [
       'article',
       {
         id: manuscriptNode.attrs.id,
+        doi: manuscriptNode.attrs.doi,
```
mbartenev-atypon, line 57: do we need that in the DOM attributes?

mbartenev-atypon (changes requested):

src/schema/nodes/affiliations.ts
mvidtman: this change is reverted....

src/schema/nodes/manuscript.ts
mvidtman: this change is reverted....

mbartenev-atypon (approved):

PR #106, opened 2024-06-23
LEAN-3577 Migration Implementation

mbartenev-atypon:


mnatsheh (commented):
I suggest we use [semver](https://www.npmjs.com/package/semver) for the versions.

package.json
```diff
@@ -34,9 +34,12 @@
     "@manuscripts/library": "1.3.7",
     "debug": "^4.3.4",
     "jszip": "^3.10.1",
+    "lodash": "^4.17.21",
+    "migration-base": "npm:@manuscripts/transform@2.3.9-LEAN-3577-support-merge",
```
mnatsheh, line 38: Can this be a dev dependency?

src/schema/migration/migrate.ts
```diff
@@ -0,0 +1,91 @@
...
+  // now find all versions that we have to migrate that do from version
+}
+
+function ensureVersionAscOrder() {
+  return migrationScripts.sort((a, b) =>
+    semver.eq(a.toVersion, b.toVersion)
+      ? 0
+      : semver.gt(a.toVersion, b.toVersion)
+      ? 1
+      : -1
```
mnatsheh, line 71: There should be a compare function in semver.

mbartenev-atypon: Indeed. Replaced.

package.json
mbartenev-atypon: Yes, we need it only for running tests. Moved.

mnatsheh (approved):

PR #113, opened 2024-07-09
LEAN-3753 -- Remove "comments" attr from PM schema

hhusban:


src/jats/__tests__/jats-roundtrip.test.ts
```diff
@@ -117,6 +118,7 @@ const roundtrip = async (filename: string) => {
   const doc = new DOMParser().parseFromString(input, 'application/xml')
 
   const models = parseJATSArticle(doc)
+  fs.writeFileSync(__dirname + '/models.json', JSON.stringify(models))
```
mnatsheh, line 121: Is this needed?

mnatsheh (approved):

src/jats/__tests__/jats-roundtrip.test.ts
hhusban: Oh, forgot to remove these

asouqi (approved):
when you open PR in `article-editor` let me know, we need to do a cleanup there also to not save comments attribute

PR #116, opened 2024-07-17
LEAN-3577 Exposing testDoc for future merge

mbartenev-atypon:
This has to be merged first as implementation in LEAN-3577 relies on 'createTestDoc' to be exposed from a testing base which is any previous package


mnatsheh (approved):
Is this going to include the test helper in the package?

mbartenev-atypon:
> Is this going to include the test helper in the package?

Yes it will be now shipped, however it's not going to be a part of any bundle as we only use it in testing for the transform.

PR #117, opened 2024-07-31
Supporting usage of ManuscriptNode instead of MPManuscript on FE

mbartenev-atypon:
- Removing manuscript custom labels for objects
- Extending ManuscriptNode attributes

### DIFF
--- a/src/transformer/html.ts
+++ b/src/transformer/html.ts
@@ -120,7 +120,7 @@ export class HTMLTransformer {
 
     this.modelMap = modelMap
     const manuscript = findManuscript(this.modelMap)
-    this.labelTargets = buildTargets(fragment, manuscript)
+    this.labelTargets = buildTargets(fragment)
     this.document = document.implementation.createDocument(
       'http://www.w3.org/1999/xhtml',
       'html',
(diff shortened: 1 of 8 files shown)


src/transformer/html.ts
```diff
@@ -119,8 +119,7 @@ export class HTMLTransformer {
     const { idGenerator, mediaPathGenerator } = options
 
     this.modelMap = modelMap
-    const manuscript = findManuscript(this.modelMap)
-    this.labelTargets = buildTargets(fragment, manuscript)
+    this.labelTargets = buildTargets(fragment)
```
mbartenev-atypon, line 122: @mnatsheh, I removed manuscript here as the second argument because this is the only reason we are passing MPManuscript to body-editor from article editor and the last reason we need MPManuscript on FE. However, I don't think we actually use labels customisation on FE. Please advise.

mnatsheh: I'm OK with removing this.

mbartenev-atypon:
This can be reviewed now.

mnatsheh (approved):

PR #123, opened 2024-08-25
LEAN-3911: remove ModelMap usages 

mbetamony:
### Changes to import: 

Old import used to do the following: 
1. parse the front into `models`
2. parse the body into a `node` using `jats-body-dom-parser`
3. parse references and comments into `models`
4. encode the the body node into `models`
5. decode all the models into a manuscript node

New import:

Now we format the content we need from the article, and run it through the [jats-dom-parser](https://github.com/Atypon-OpenSource/manuscripts-transform/blob/de649a9e9f9ff7b4719681950fce6f79ede0ea61/src/jats/importer/jats-dom-parser.ts), preivously `jats-body-dom-parser`,

We now only need the decoder for `jats-export` on [the api](https://github.com/Atypon-OpenSource/manuscripts-api/blob/35398d750e4e6483f7a238c21251fabc9557b385/src/DomainServices/ProjectService.ts#L416), technically the model map is our only source of truth as the document is always being updated. We can solve that by creating an originalDoc to use as a source of truth, and get rid of the decoder. (let me know your thoughts)

### Changes to export: 
I kept things simple, instead of looking for a specific model we look for the opposing node and fetch the content. I think ideally we'd use our serializer to transform all nodes, and then move the content as we like, but i wanted to keep it simple for now so we can easily debug it in case any errors are reported. 

Please ignore the tests for now, i'll update these soon.


src/jats/importer/create-article-node.ts
```diff
@@ -0,0 +1,45 @@
...
+    }
+  })
+
+  const article = document.createElement('article')
+  const title = document.createElement('article-title')
+  article.appendChild(manuscriptEl)
+  article.appendChild(title)
+
+  return jatsBodyDOMParser.parse(article).firstChild as ManuscriptNode
+}
```
mbetamony, line 45: this is used to generate our node in case we don't have JATS.

src/jats/importer/jats-comments.ts
```diff
@@ -50,175 +37,60 @@ export const isJATSComment = (node: Node) => {
...
-      comments.push(...processModel(model, marks))
-    } else if (isKeyword(model)) {
-      comments.push(...processKeyword(model, marks))
-    }
-  }
-  return comments
+const createHighlightMarkerElement = (doc: Document, id: string) => {
+  const highlightMarker = doc.createElement('highlight-marker')
+  highlightMarker.setAttribute('id', id)
+  highlightMarker.setAttribute('position', 'point')
```
mbetamony, line 83: Debugging this i only found that we use `point` comments on the import. Please correct me if I'm mistaken

src/jats/jats-exporter.ts
```diff
@@ -110,28 +96,36 @@ const findChildNodeOfType = (
...
-      'vocab-identifier': CREDIT_VOCAB_IDENTIFIER,
-      'vocab-term': role.name,
-      'vocab-term-identifier': role.uri,
-    }
-  }
-
-  return {
-    vocab: 'uncontrolled',
-  }
+const findNodesOfType = (
```
mnatsheh, line 99: Can we use the prosemirror-utils package instead of implementing it ourselves? Also do we need to send to send the fragment? it looks like we always use `this.fragment`.

src/jats/jats-exporter.ts
```diff
@@ -110,28 +96,36 @@ const findChildNodeOfType = (
...
+  fragment.descendants((node) => {
+    if (node.type === nodeType) {
+      nodes.push(node)
+    }
+    return recurse
+  })
+  return nodes
 }
 
+const findNode = (
```
mnatsheh, line 114: Same as above.

src/jats/jats-exporter.ts
```diff
@@ -424,10 +433,14 @@ export class JATSExporter {
     }
   }
 
-  protected buildFront = (doi?: string, id?: string, links?: Links) => {
-    const manuscript = findManuscript(this.modelMap)
-    const titles = findTitles(this.modelMap)
-
+  protected buildFront = (
+    manuscript: Manuscript,
```
mnatsheh, line 437: Do we actually need this? We should be able to derive any information directly from the manuscript node.

src/jats/jats-exporter.ts
```diff
@@ -424,10 +433,14 @@ export class JATSExporter {
...
+    journal?: Journal,
+    doi?: string,
+    id?: string,
+    links?: Links
```
mnatsheh, line 441: Are these used? We should just use whatever is in the manuscript node.

src/jats/jats-exporter.ts
```diff
@@ -525,52 +534,56 @@ export class JATSExporter {
...
+    //   element.setAttribute('alt-title-type', 'right-running')
+    //   this.setTitleContent(element, titles.runningTitle)
+    //   titleGroup.appendChild(element)
+    // }
+    const supplementsNodes = findNodesOfType(
+      this.fragment,
+      schema.nodes.supplement
+    )
+    supplementsNodes.forEach((node) => {
+      if (node.type !== schema.nodes.section_title) {
```
mnatsheh, line 560: Is this check needed? You're looking up the nodes by type already.

src/jats/jats-exporter.ts
```diff
@@ -969,19 +982,19 @@ export class JATSExporter {
 
         const text = cross.attrs.customLabel || cross.attrs.label
 
-        const model = getModel(rids[0])
-        if (!model) {
+        const crossRef = findNode(this.fragment, rids[0])
```
mnatsheh, line 985: Let's call this "target".

src/jats/jats-exporter.ts
mnatsheh: This also applies to all the options. Anything we don't use in pressroom we can drop.

src/jats/jats-exporter.ts
```diff
@@ -525,52 +534,56 @@ export class JATSExporter {
...
-      (model) => model.objectType === ObjectTypes.Supplement
-    ) as Supplement[] | undefined
-    if (supplements && supplements.length > 0) {
-      for (const supplement of supplements) {
+    // if (titles.runningTitle) {
+    //   const element = this.document.createElement('alt-title')
+    //   element.setAttribute('alt-title-type', 'right-running')
+    //   this.setTitleContent(element, titles.runningTitle)
+    //   titleGroup.appendChild(element)
+    // }
```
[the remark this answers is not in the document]
mnatsheh: We can remove this for now.

mnatsheh:
Regarding imports, can't we just replace json-schema types with their equivalent PM attrs type without extensive changes? e.g. replace `BibliographyItem` with the `Attrs` type in `BibliographyNode`.

mbetamony:
> Regarding imports, can't we just replace json-schema types with their equivalent PM attrs type without extensive changes? e.g. replace `BibliographyItem` with the `Attrs` type in `BibliographyNode`.

Just replace the types right? I think this applies to the manuscirpt, journal, and references models. I can change them but I'd like to confirm if that's what you meant.

src/jats/jats-exporter.ts
mbetamony: No, you're right.

src/jats/jats-exporter.ts
mbetamony: There are attributes that are available in the model but missing in the manuscript node, e.g. `figureCount`, `acceptanceDate`, and many more. We can update the manuscript node to have these attributes.

src/jats/jats-exporter.ts
mbetamony: Sure, i'll use the porsemirror-utils package.

src/jats/jats-exporter.ts
mnatsheh: We should be able to derive this info from the document, right? No need to store it explicitly.

mnatsheh:
> > Regarding imports, can't we just replace json-schema types with their equivalent PM attrs type without extensive changes? e.g. replace `BibliographyItem` with the `Attrs` type in `BibliographyNode`.
> 
> Just replace the types right and keep the original approach (format the article and send it through the parser)? Or do you want us to go with a different approach and generate the nodes directly?

Right, can we keep the logic and change the types?

mbetamony:
I did recheck the comments and highlight nodes, and they do share the same ids in the older implementation.

src/schema/nodes/bibliography_item.ts
```diff
@@ -14,16 +14,32 @@
...
+}
+export interface BibliographyItemDate {
+  'date-parts'?: [
+    [LooseNumber, LooseNumber?, LooseNumber?],
+    [LooseNumber, LooseNumber?, LooseNumber?]?
+  ]
+  _id: string
+  objectType: 'MPBibliographicDate'
+}
+export interface BibliographyItemAttributes {
```
mnatsheh, line 38: Using the json-schema types should be OK. No need to define our own.

src/jats/importer/create-article-node.ts
```diff
@@ -0,0 +1,37 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Manuscript } from '@manuscripts/json-schema'
+
+import { defaultTitle } from '../../lib/deafults'
+import { ManuscriptNode, schema } from '../../schema'
+
+export const createArticleNode = (manuscript: Partial<Manuscript>) => {
```
mnatsheh, line 22: We should accept `Partial<ManuscriptAttrs>` instead of the json-schema type, where `ManuscriptAttrs`  is the `Attrs` defined inside `manuscripts.ts`.

src/jats/importer/jats-front-transformations.ts
```diff
@@ -0,0 +1,95 @@
...
+    } else {
+      title = createElement('article-title')
+      title.innerHTML = defaultTitle
+    }
+    return title
+  },
+  createAuthorNotes(document: Document) {
+    const authorNotes = document.querySelector('article-meta > author-notes')
+    if (authorNotes) {
+      const sectionTitle = document.createElement('title')
```
mnatsheh, line 59: We should send `createElement` .

src/schema/nodes/bibliography_item.ts
mnatsheh, line 38: Let's call this `BibliographyItemAttrs` for consistency with similar things done in body-editor.

mbetamony: Sure, i replaced them with the json-schema types

src/jats/importer/create-article-node.ts
mbetamony: Done.

src/jats/importer/jats-front-transformations.ts
mbetamony: Done.

src/schema/nodes/bibliography_item.ts
mbetamony: Sure, done.

mnatsheh:
The changes look generally OK, I'm guessing you're working on the tests.

mbetamony:
> The changes look generally OK, I'm guessing you're working on the tests.

Yeah, but i might move this to QA before finishing them up to speed things up if that's okay.

src/jats/importer/jats-body-transformations.ts
```diff
@@ -415,4 +417,91 @@ export const jatsBodyTransformations = {
...
+      const bibliographyItem = this.createBibliographyItem(item, createElement)
+      bibliographyElement.appendChild(bibliographyItem)
+    })
+
+    bibliographySection.appendChild(bibliographyElement)
+    backmatter.appendChild(bibliographySection)
+  },
+  createBibliographySection(createElement: (tagName: string) => HTMLElement) {
+    const section = createElement('sec')
+    section.setAttribute('sec-type', 'bibliography_section')
```
mnatsheh, line 444: We should change this to `bibliography`, otherwise it won't be detected correctly.

src/jats/importer/jats-dom-parser.ts
```diff
@@ -111,6 +128,227 @@ const marks: MarkRule[] = [
...
+          }
+        }
+        return { department, institution }
+      }
+      const getAddressLine = (element: HTMLElement, index: number) => {
+        return (
+          getTrimmedTextContent(element, `addr-line:nth-of-type(${index})`) ||
+          ''
+        )
+      }
```
mnatsheh, line 330: Can we move these outside the `getAttrs`  function?

src/jats/importer/jats-dom-parser.ts
```diff
@@ -461,6 +693,54 @@ const nodes: NodeRule[] = [
     tag: 'sec[sec-type="backmatter"]',
     node: 'backmatter',
   },
+  {
+    tag: 'bibliography-element',
```
mnatsheh, line 697: We need to also add a rule for `bibliography_section` node.

src/jats/importer/jats-reference-parser.ts
```diff
@@ -125,59 +107,47 @@ export const jatsReferenceParser = {
       const year = getTrimmedTextContent(element, 'year')
 
       if (year) {
-        bibliographyItem.issued = buildBibliographicDate({
+        bibliographyItem.issued = {
```
mnatsheh, line 110: Can we continue using `buildBibliographicDate`?

src/jats/importer/jats-reference-parser.ts
```diff
@@ -125,59 +107,47 @@ export const jatsReferenceParser = {
...
-      // TODO: handle missing person-group-type?
-      // TODO: handle contrib-group nested inside collab
-      // TODO: handle collab name
-
       const authors: BibliographicName[] = []
 
       authorNodes.forEach((authorNode) => {
-        const name = buildBibliographicName({})
-
+        const name: BibliographicName = {
```
mnatsheh, line 126: Can we continue using `buildBibliographicName`?

src/lib/utils.ts
```diff
@@ -96,3 +96,20 @@ export const getTrimmedTextContent = (
   }
   return node.querySelector(querySelector)?.textContent?.trim()
 }
+
+export const timestamp = () => Math.floor(Date.now() / 1000)
+
+export const generateAttachmentFilename = (
```
mnatsheh, line 102: We should get rid of the attachments code from the exporter. We can then move this under tests.

src/jats/importer/jats-reference-parser.ts
mbetamony: Sure.

src/jats/importer/jats-reference-parser.ts
mbetamony: Sure.

src/jats/importer/jats-body-transformations.ts
mbetamony: Good catch, thanks, i'll update it.

src/jats/importer/jats-dom-parser.ts
mbetamony: Sure

src/lib/utils.ts
mbetamony: I removed the attachment code, please let me know if we need to remove more things from there.
https://github.com/Atypon-OpenSource/manuscripts-transform/pull/123/commits/ecbb375f7729072b23b01c35938e36dccf66badd#diff-ffda4a3ab811824dfbdc9fe411843d37544de380122482986f6d3187417c68ba

src/jats/importer/jats-dom-parser.ts
mbetamony: Done.

src/jats/exporter/jats-exporter.ts
```diff
@@ -154,7 +154,7 @@ export type CSLOptions = {
 export type ExportOptions = {
   version?: Version
   journal?: Journal
-  idGenerator: IDGenerator,
+  idGenerator: IDGenerator
```
mbetamony, line 157: We don't send this from the api, is it needed here?

src/jats/importer/jats-comments.ts
```diff
@@ -18,56 +18,48 @@ import { ObjectTypes } from '@manuscripts/json-schema'
...
-          node.parentNode?.insertBefore(highlightMarker, node)
-          const commentElement = createCommentElement(doc, comment)
-          commentsElement.appendChild(commentElement)
+        const text = parseJATSComment(node)
+        if (text) {
+          const id = generateID(ObjectTypes.CommentAnnotation)
+          const parent = node.parentNode as Element
+          if (parent) {
+            //todo
+            const marker = createHighlightMarkerElement(doc, id)
```
mbetamony, line 53: Is there something left to do here?

src/jats/importer/jats-body-transformations.ts
```diff
@@ -14,13 +14,17 @@
  * limitations under the License.
  */
```
mbetamony, line 16: Should we call this file jats-transformations?

src/jats/importer/jats-comments.ts
```diff
@@ -14,211 +14,80 @@
...
 }
 
-/**
- * Replaces processing instructions with tokens
- */
-export const markComments = (doc: Document): JATSCommentMark[] => {
-  const marks: JATSCommentMark[] = []
-  const root = doc.getRootNode()
-  const queue: Node[] = [root]
+export const markComments = (doc: Document) => {
```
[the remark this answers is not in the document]
mnatsheh: The problem is at this stage most nodes might not have ids (e.g. paragraphs). That's why the `updateCommentTargets` function was needed, to inject the target ids for comments without a target id.

mnatsheh (approved):

mnatsheh:
I feel this deserves a major release.

mbetamony:
> I feel this deserves a major release.

Sure, that feels right. I think Zax would merge all the PRs as major releases, so i'll do that manually.

PR #131, opened 2024-09-04
LEAN-3880 -- boxed-text schema

hhusban:


src/schema/nodes/box_element.ts
```diff
@@ -0,0 +1,67 @@
...
+  attrs: {
+    id: { default: '' },
+    label: { default: '' },
+    dataTracked: { default: null },
+  },
+  group: 'block element',
+  selectable: false,
+  parseDOM: [
+    {
+      tag: 'boxed-text',
```
mnatsheh, line 41: I think this should be HTML, so something like "div.box".

src/schema/nodes/box_element.ts
```diff
@@ -0,0 +1,67 @@
...
+  id: string
+  label: string | null
+}
+
+export interface BoxElementNode extends ManuscriptNode {
+  attrs: Attrs
+}
+
+export const box_element: NodeSpec = {
+  content: 'section_label? figcaption? section?',
```
mnatsheh, line 31: Shouldn't this be in the attrs?

src/schema/nodes/section.ts
```diff
@@ -56,7 +56,8 @@ const choosePageBreakStyle = (element: HTMLElement): number => {
 
 export const section: NodeSpec = {
   // NOTE: the schema needs paragraphs to be the default type, so they must explicitly come first
-  content: 'section_label? section_title (paragraph | element)* sections*',
+  content:
+    'section_label? section_title (box_element | paragraph | element)* sections*',
```
mnatsheh, line 60: We should not need to update this, since the box_element is an "element".

mnatsheh (approved):

src/schema/nodes/box_element.ts
hhusban: Since this is treated as a section and the label will already handled by the current section_label, this will minimize the changes, and I removed the label from the schema attributes.

PR #132, opened 2024-09-05
LEAN-3881 -- box-element importer and exporter

hhusban:


asouqi:
@hhusban can you please add `[schema.nodes.box_element, 'Box']` to [nodeNames](https://github.com/Atypon-OpenSource/manuscripts-transform/blob/master/src/transformer/node-names.ts#L20) and type in [labelledNodeTypes](https://github.com/Atypon-OpenSource/manuscripts-transform/blob/7522ebc46d203705056a9786273a77a735d47916/src/transformer/labels.ts#L36)

PR #138, opened 2024-09-17
LEAN-3882: funder group schema

mbetamony:


mnatsheh (changes requested):
My suggestion is to go with the following:

```
manuscript
  awards
    award
      id
      code
      recipient
      source
```


JATS4R indicates that only a single "funding-group" should be there, so we can ignore it for import. Each "award-group" would be mapped to an "award".

mbetamony:
> My suggestion is to go with the following:
> 
> ```
> manuscript
>   awards
>     award
>       id
>       code
>       recipient
>       source
> ```
> 
> JATS4R indicates that only a single should be there, so we can ignore it for import. Each would be mapped to an "award".

A single what?

mnatsheh:
> > My suggestion is to go with the following:
> > ```
> > manuscript
> >   awards
> >     award
> >       id
> >       code
> >       recipient
> >       source
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > JATS4R indicates that only a single should be there, so we can ignore it for import. Each would be mapped to an "award".
> 
> A single what?

Sorry, some problem with GitHub rendering.

mbetamony:
> My suggestion is to go with the following:
> 
> ```
> manuscript
>   awards
>     award
>       id
>       code
>       recipient
>       source
> ```
> 

Isn't what is suggested here the same as what is implemented, just with different naming(apart from the awardIDs, we can have it as a single string and split it)? Unless you want what's under award to be nodes instead of attributes.

> JATS4R indicates that only a single "funding-group" should be there, so we can ignore it for import. Each would be mapped to an "award".

Can you please elaborate on this? why would we ignore it on import? It feels like I'm missing something here

> Each would be mapped to an "award".

Each of  what exactly? We can only have 1 funding group that can have multiple awards, no?

mnatsheh:
What I mean is that we do _not_ have to have a 1-to-1 mapping between "funding-group" in the input JATS and the PM dom (which is why I suggest calling the group node "awards" instead of "fundingGroup").

mbetamony:
> What I mean is that we do _not_ have to have a 1-to-1 mapping between "funding-group" in the input JATS and the PM dom (which is why I suggest calling the group node "awards" instead of "fundingGroup"). What we care about are "award-group"s.

I renamed the nodes, please let me know if that's what you meant.

mnatsheh (approved):
Looks good. Please update the "manuscript" node schema as well.

mbetamony:
> Looks good. Please update the "manuscript" node schema as well.

Should I add it right after the title? or somewhere after?

mnatsheh:
> > Looks good. Please update the "manuscript" node schema as well.
> 
> Should I add it right after the title? or somewhere after?

Let's put it after author_notes.

PR #145, opened 2024-09-30
LEAN-4011 -- Store manuscript dates in the PM doc

hhusban:


hhusban:
@mnatsheh something is a bit weird here, the dom-parser never actually went through the "body" tag, when I added "article" wrapper (used article since it makes sense as a wrapper), now it actually uses the "body" mark

mnatsheh:
Let's wait until LEAN-3911 is merged, as it will change how import/export work.

hhusban:
@mbetamony this file "src/jats/importer/jats-body-transformations.ts" seems to be unused, should we remove it?

mbetamony:
> @mbetamony this file "src/jats/importer/jats-body-transformations.ts" seems to be unused, should we remove it?

Yes please, I thought i removed it in LEAN-3911.

src/jats/importer/jats-transformations.ts
```diff
@@ -477,6 +477,13 @@ export const moveReferencesToBackmatter = (
   backmatter.appendChild(section)
 }
 
+export const moveHistoryToBody = (document: Document, body: Element) => {
```
mnatsheh, line 480: Do we need to move the dates? can we read them directly from their original location?

src/schema/nodes/manuscript.ts
```diff
@@ -28,6 +28,14 @@ export interface ManuscriptAttrs {
   articleType: string
   prototype: string
   primaryLanguageCode: string
+  dates: {
```
mnatsheh, line 31: I don't think we need the `dates` wrapper, we can just have the dates as a flat list under attrs.

mnatsheh (changes requested):

src/jats/importer/jats-transformations.ts
hhusban: Prior to the changes in LEAN-3911 it was necessary, but it is no longer needed, so i updated the PR

src/schema/nodes/manuscript.ts
hhusban: updated

mnatsheh (approved):

mbetamony:
We'll also need to make changes to the export; they were removed in LEAN-3911 because we stopped sending the manuscript model to the export. 

I'll push the updates here if that's OK; since there's a [P1 ticket](https://jira.atypon.com/browse/LEAN-4048) filed, please feel free to make any changes

mbetamony:
I also wonder if we need to update the manuscript model with these dates when importing the jats; we used to previously store them there. But since we're moving away from json schema i'm assuming we don't need to, correct?

mnatsheh:
> We'll also need to make changes to the export; they were removed in LEAN-3911 because we stopped sending the manuscript model to the export.
> 
> I'll push the updates here if that's OK; since there's a [P1 ticket](https://jira.atypon.com/browse/LEAN-4048) filed, please feel free to make any changes

Good catch.

> I also wonder if we need to update the manuscript model with these dates when importing the jats; we used to previously store them there. But since we're moving away from json schema i'm assuming we don't need to, correct?

We shouldn't need to.

PR #149, opened 2024-10-09
LEAN-3933 LW-FY25-Q2-S2: No default category is set for the sub-section

fwaisi:


src/jats/__tests__/__snapshots__/jats-importer.test.ts.snap
```diff
@@ -17912,7 +17912,7 @@ Object {
             },
             Object {
               "attrs": Object {
-                "category": "MPSectionCategory:materials-method",
+                "category": "MPSectionCategory:subsection",
```
mbetamony, line 17915: Are you sure this is the correct behavior?

fwaisi: I have tested the changes, and everything is working just fine. Regarding the previous case, it is indeed correct since it’s a subsection.

src/transformer/section-category.ts
```diff
@@ -328,8 +330,11 @@ export const chooseSectionCategory = (
 ): SectionCategory | undefined => {
   const secType = section.getAttribute('sec-type') as SecType
   const secCat = chooseSectionCategoryByType(secType)
+  const parent = section.parentNode?.parentNode?.nodeName.toLowerCase()
   if (secCat) {
     return secCat
+  } else if (parent !== 'body') {
+    return 'MPSectionCategory:subsection'
   } else {
```
mbetamony, line 338: can we have this logic in either the `jats-dom-parser.ts` when handling the section tag, or even before?

src/jats/importer/jats-dom-parser.ts
```diff
@@ -863,7 +863,13 @@ const nodes: NodeRule[] = [
...
     getAttrs: (node) => {
       const element = node as HTMLElement
-
+      const grandParentNodeName =  element.parentNode?.parentNode?.nodeName.toLowerCase()
+      if ( grandParentNodeName !== 'body' ) {
+        return {
+          id: element.getAttribute('id'),
+          category: 'MPSectionCategory:subsection',
+        }
+      }
```
mbetamony, line 872: We can just set the sec-type attr to 'subsection', we don't need to explicitly return this.
 `element.setAttribute('sec-type', 'subsection')`

mbetamony (approved):

src/jats/importer/jats-dom-parser.ts
mbetamony: thanks, another question is should we check if the grandParentNodeName is defined?  What if the node doesn't have a grand parent? The current implementation would set it as a subsection, no?

fwaisi: yes, fixed.

PR #152, opened 2024-10-14
LEAN-4053 - remove inline_footnote contents

mnatsheh:


src/schema/nodes/inline_footnote.ts
```diff
@@ -31,7 +30,6 @@ export const inlineFootnote: NodeSpec = {
   name: 'Footnote marker',
   attrs: {
     rids: { default: [] },
-    contents: { default: '' },
```
mbetamony, line 34: We probably need to update the `jats-dom-parser` as well. I also wonder if we need to migrate older documents since i think we did store content there.

mbetamony (approved):

src/schema/nodes/inline_footnote.ts
mnatsheh: Updated. The value of the attr should not be relevant, so we don't need to worry about older documents.

mbetamony: Won't we need to remove the attribute as well from the node? I'm wondering what would happen if we don't.

PR #154, opened 2024-10-15
Fix date storage and format

hhusban:
in JS native date the year and day are 1 based but the month is zero based, so when parsing a month with value 12 it adds a year and set the month to 0, to fix this we make the month zero based (subtracting 1) and when exporting the date to make sure we have the correct month we add 1, I also added a format to make sure that the month is suffixed with zero when needed

@mnatsheh do we also need to format the day to have a leading zero too?


mbetamony:
A question I have about this is why do we need to transform the values to timestamps on import? Can't we just import them as strings and export them as is?

hhusban:
> A question I have about this is why do we need to transform the values to timestamps on import? Can't we just import them as strings and export them as is?

Good question, but i assumed that the editor might format the date differently from what is used in the export, and this would make it easier for FE to work with

mnatsheh (approved):
> @mnatsheh do we also need to format the day to have a leading zero too?

Let's do it for day as well.

PR #156, opened 2024-10-24
LEAN-3978 LW-FY25-Q2-S3: Table_element delete suggestion is not showi…

fwaisi:
…ng when cut the table from the content


fwaisi:
@IslamJMomani, Kindly note that I have reverted the changes here, since its not related to the bug in the ticket.

PR #158, opened 2024-10-28
LEAN-3961: consolidate section category definitions

mbetamony:
1- defined a new type for the [section-categories](https://github.com/Atypon-OpenSource/manuscripts-transform/pull/158/files#diff-bd504b993b890189fe863472a904b264c51a45cca0a7e7cac699e5f9348be749R152): 
- id
- name
- groupIDs
- isDisabled
- synonyms -> includes the different names we used in [chooseSecCategoryByType](https://github.com/Atypon-OpenSource/manuscripts-transform/blob/1467dccadced5a198f68ca31de18f31a17477a6a/src/transformer/section-category.ts#L246C14-L246C41) and [chooseSecCategoryByTitle](https://github.com/Atypon-OpenSource/manuscripts-transform/blob/1467dccadced5a198f68ca31de18f31a17477a6a/src/transformer/section-category.ts#L350)
- titles -> used for autocomplete 
- isEditable -> used in [body-editor](https://github.com/Atypon-OpenSource/manuscripts-body-editor/blob/939371c0b65077c4a2402dc33e618c0c8963aff5/src/lib/section-categories.ts#L32)

I removed all properties that weren't used.

2- renamed some sections so we don't need to handle specific cases that was done by [chooseSecType](https://github.com/Atypon-OpenSource/manuscripts-transform/blob/1467dccadced5a198f68ca31de18f31a17477a6a/src/transformer/section-category.ts#L202)

3- added the type [ManuscriptTemplate](https://github.com/Atypon-OpenSource/manuscripts-transform/pull/158/files#diff-bd504b993b890189fe863472a904b264c51a45cca0a7e7cac699e5f9348be749R162) in the transformer to support having section categories inside the template (we can probably get rid of `createdAt` and `updatedAt`, we kept those after removing manuscripts-data, not sure why)

4- this doesn't belong here but i removed [table-cell-styles.ts](https://github.com/Atypon-OpenSource/manuscripts-transform/pull/158/files#diff-cdfb14b87d6d9db9de31251d8a302378adca060cb6c2f90df16392f61988160b), it doesn't seem to be used anywhere


src/schema/types.ts
```diff
@@ -147,3 +148,24 @@ export type DataTrackedAttrs = {
...
 }
+
+export type SectionCategory = {
+  _id: string
+  name: string
+  synonyms: string[]
+  isEditable: boolean
+  isDisabled?: boolean
+  titles?: string[]
+  groupIDs?: string[]
```
mnatsheh, line 159: The idea with `groupIDs` never materialized, so I think we should revisit it. Maybe we can get rid of it completely, and update the template to have a separate list for body section categories and backmatter section categories?

src/schema/types.ts
```diff
@@ -147,3 +148,24 @@ export type DataTrackedAttrs = {
...
   createdAt: number
 }
+
+export type SectionCategory = {
+  _id: string
+  name: string
+  synonyms: string[]
+  isEditable: boolean
+  isDisabled?: boolean
+  titles?: string[]
```
mnatsheh, line 158: Can we really have a category without a title?

src/schema/types.ts
```diff
@@ -147,3 +148,24 @@ export type DataTrackedAttrs = {
   userID: string
   createdAt: number
 }
+
+export type SectionCategory = {
+  _id: string
+  name: string
```
mnatsheh, line 154: I don't think this is needed, the "titles" should be enough.

src/schema/types.ts
mbetamony: I'm using the titles property for auto-complete (which only a few sections use). I had to differentiate between titles and synonyms because the old titles we had weren't the ones used by the autocomplete but were used for the import and I didn't think we want to support all previous titles for autocomplete.

mbetamony: If we get rid of the name, then can probably have titles as required and use the first title as the name.

src/jats/exporter/jats-exporter.ts
```diff
@@ -2118,7 +2114,10 @@ export class JATSExporter {
     footnotes.forEach((fn) => {
       const fnType = fn.getAttribute('fn-type')
       if (fnType) {
-        fn.setAttribute('fn-type', chooseJatsFnType(fnType))
+        fn.setAttribute(
```
mnatsheh, line 2117: what's the purpose of this? why is the "fn-type" not correct from the start?

src/schema/types.ts
mbetamony: So, you mean we'd have 4 lists? abstracts, body, backmatter and other sections? And in the [body-editor](https://github.com/Atypon-OpenSource/manuscripts-body-editor/blob/6d24b2031be40663f75b20d2f03d089ca3ac5582/src/plugins/section-category/section-category-utils.ts#L209) we'd check if the section belongs to a certain list?

mbetamony: It also seems to me that the purpose of groupIDs being an array is lost, the usages assume that we can only have one group id for a section. Can we not have a section that can have multiple group ids?

mbetamony:
It seems that even though we have an `abstracts` node, the way its handled is that we have an abstract node that will have an abstracts section (section node) inside it, so we might want to need to keep the 'abstracts' category (though this might require us to have the 'abstracts' back as a section group). Other thing might be we change the way we import the abstracts; it seems that we expect different types of abstract sections, and we add them to the abstract node (check [moveAbstracts](https://github.com/Atypon-OpenSource/manuscripts-transform/blob/b99ab1a7f48dc941f385fd3401af18ecb6ccc205/src/jats/importer/jats-transformations.ts#L129))

mnatsheh:
> Further, the History tab provides access to Snapshots, so in those other apps where the Editor is used, if I'm a Viewer would the approach suggested here mean that I'd sometimes have access to snapshots (if current assignee had made a change) and sometimes not have access to snapshots (if current assignee has not yet made a change... or they have made changes, then accepted them)?


This is a transient operation though, we still parse that into an abstracts node in the end.

mnatsheh:
> And about changing the templates to have the csl id instead of the bundle, I'm a bit hesitant since this won't be backwards compatible with the WM, we'd probably need to update it, let me know your thoughts.

That's fine let's keep it for now.

src/lib/sectionCategories.ts
```diff
@@ -0,0 +1,25 @@
+/*!
```
mnatsheh, line 1: Let's call this `section-categories.ts`.

src/schema/migration/migration-scripts/3.0.12.ts
```diff
@@ -0,0 +1,47 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { JSONNode } from '../migrate'
+import { MigrationScript } from '../migration-script'
+
+export class Migration3010 implements MigrationScript {
```
mnatsheh, line 19: You need to add this to the list in `index.ts` for it to take effect I think.

src/schema/types.ts
```diff
@@ -147,3 +148,27 @@ export type DataTrackedAttrs = {
   userID: string
   createdAt: number
 }
+
+export enum SectionGroup {
```
mnatsheh, line 152: It might be simpler to just use a union type:

`export type SectionGroup = 'body' | 'backmatter'`

src/jats/__tests__/__snapshots__/jats-importer.test.ts.snap
```diff
@@ -796,7 +796,7 @@ Object {
       "content": Array [
         Object {
           "attrs": Object {
-            "category": "MPSectionCategory:abstract",
+            "category": "",
             "dataTracked": null,
             "id": "MPSection:test",
           },
```
mbetamony, line 802: @mnatsheh, this is what I meant by my [comment](https://github.com/Atypon-OpenSource/manuscripts-transform/pull/158#issuecomment-2450539295), the sections under the `abstracts` node are added into a section called abstract. for example, this is what the dom-parser receives from [jats-import.xml](https://github.com/Atypon-OpenSource/manuscripts-transform/blob/master/src/jats/__tests__/__fixtures__/jats-import.xml)
```
<sec sec-type="abstracts"><title>Abstracts</title>
  <sec sec-type="abstract"><title>Abstract</title>
     *** other sections, e.g Objectives
```
I'm not sure if that behavior is intended or we need to fix it.

src/jats/__tests__/__snapshots__/jats-importer.test.ts.snap
```diff
@@ -1284,7 +1284,7 @@ Object {
         },
         Object {
           "attrs": Object {
-            "category": "MPSectionCategory:materials-method",
```
mbetamony, line 1287: Are we sure we're not importing this section category anymore?

src/jats/__tests__/data/section-categories.ts
```diff
@@ -0,0 +1,227 @@
...
+  },
+  {
+    id: 'abstract',
+    titles: ['Abstract'],
+    synonyms: ['abstract'],
+    group: 'abstracts',
+    isUnique: false,
+  },
+  {
+    id: 'abstract-graphical',
```
mnatsheh, line 35: We should not need this since this has its own node.

mbetamony: Removed.

src/schema/types.ts
mbetamony: Done.

src/lib/sectionCategories.ts
mbetamony: Done.

src/jats/__tests__/section-categories.json
```diff
@@ -0,0 +1,266 @@
+[
+  {
+    "_id": "MPSectionCategory:abbreviations",
```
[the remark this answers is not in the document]
mbetamony: This is done.

src/schema/types.ts
mbetamony: Removed.

src/schema/types.ts
```diff
@@ -147,3 +148,24 @@ export type DataTrackedAttrs = {
   userID: string
   createdAt: number
 }
+
+export type SectionCategory = {
+  _id: string
```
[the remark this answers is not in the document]
mbetamony: Changed.

mbetamony:
The reason for changing the name of the `introduction` section to `intro` is because we export it as `intro`. (Same for other sections in this [function](https://github.com/Atypon-OpenSource/manuscripts-transform/blob/1467dccadced5a198f68ca31de18f31a17477a6a/src/transformer/section-category.ts#L202))

src/__tests__/data/project-dump-2.json
mbartenev-atypon, line 1: were we not using any of those?

mnatsheh: These are json-schema files, which we should not be using.

src/jats/__tests__/data/section-categories.ts
```diff
@@ -0,0 +1,220 @@
...
+    group: 'body',
+    isUnique: false,
+  },
+  {
+    id: 'subsection',
+    titles: ['Subsection'],
+    synonyms: ['subsection'],
+    group: 'body',
+    isUnique: false,
+  },
```
mnatsheh, line 205: We can get rid of these now.

src/lib/section-categories.ts
```diff
@@ -14,24 +14,12 @@
...
  */
 
-export type SectionGroupTypeID = 'abstracts' | 'body' | 'backmatter'
+import { SectionCategory, SectionGroup } from '../schema'
 
-export type SectionGroupType = {
-  _id: SectionGroupTypeID
-  title: string
-}
+export const isSubsection = (categoryID: string) => categoryID === 'subsection'
```
mnatsheh, line 19: We can get rid of this.

mnatsheh (commented):
Per discussion with Fahid, we can get rid of the "subsection" category and its usages. We should also update the dom-parser to no inject that sec-type when importing.

mbetamony:
> Per discussion with Fahid, we can get rid of the "subsection" category and its usages. We should also update the dom-parser to no inject that sec-type when importing.

Removed, and I updated the migration to remove the sec-type `subsection`.

PR #159, opened 2024-11-06
LEAN-4066: fix the import of graphical abstract section

mbetamony:
One question i have is, how do we usually import the graphical abstract? do we expect it it in the xml, or is it something that literatum adds if it finds an attachment with the type `GRAPHICAL-ABSTRACT-IMAGE`. (and what if we do receive one in the xml and one in the attachment? e.g [LEAN-4067](https://jira.atypon.com/browse/LEAN-4067))


mnatsheh:
This comes from the time when the source document is not JATS (i.e. docx). When we convert that to JATS, the output doesn't include GA tagging, so added the "GRAPHICAL-ABSTRACT-IMAGE" file type so that WM can add the tagging. I believe that when JATS is the source, we should tag GAs directly and not depend on WM, I will talk to QA about it.

Do you know why this "regressed"? are we sure it worked before?

src/jats/__tests__/__fixtures__/jats-import.xml
```diff
@@ -152,6 +152,12 @@
             may postpone recurrent stroke and death in stroke survivors.</p>
         </sec>
       </abstract>
+      <abstract abstract-type="graphical">
+        <fig id="fig-1" position="anchor" fig-type="figure">
+          <label>Figure 1</label>
```
mnatsheh, line 157: GAs should not have labels.

mbetamony:
> This comes from the time when the source document is not JATS (i.e. docx). When we convert that to JATS, the output doesn't include GA tagging, so added the "GRAPHICAL-ABSTRACT-IMAGE" file type so that WM can add the tagging. I believe that when JATS is the source, we should tag GAs directly and not depend on WM, I will talk to QA about it.
> 
> Do you know why this "regressed"? are we sure it worked before?

Yes, before removing the modelmap usages, the dom parser detected this as a section (not a GA section, similar to what's happening now), but the [decoder](https://github.com/Atypon-OpenSource/manuscripts-transform/blob/01eba7561b5e0d9e6a3709d1af8e7889901af44e/src/transformer/decode.ts#L784) took care of making it a GA section.

src/jats/__tests__/__fixtures__/jats-import.xml
mbetamony: Removed.

PR #161, opened 2024-11-11
LEAN-4095:  Support the @fig-type attribute for the <fig> element

mbetamony:


src/schema/nodes/figure_element.ts
```diff
@@ -35,6 +38,7 @@ export const figureElement: NodeSpec = {
     id: { default: '' },
     attribution: { default: undefined },
     dataTracked: { default: null },
+    position: { default: undefined },
```
mnatsheh, line 41: Based on my comment in the ticket, let's change this to "type".

mbetamony: We currently export the figure node as a figure element with  `fig-type = figure`, and the  listing node as a figure element with the `fig-type = listing`. We're not going to do that anymore?

mnatsheh: Yeah we're likely going to stop that. We're getting rid of listings at some point, so how we export them now doesn't matter much.

mbetamony: Got it, done.

mnatsheh (approved):

PR #164, opened 2024-11-20
LEAN-4149: import/export/schema for graphic

mbetamony:


src/schema/nodes/media.ts
```diff
@@ -0,0 +1,55 @@
...
+}
+
+export const media: NodeSpec = {
+  attrs: {
+    id: { default: '' },
+    dataTracked: { default: null },
+    href: { default: '' },
+    show: { default: '' },
+    mimetype: { default: '' },
+    mimeSubtype: { default: '' },
```
mbetamony, line 40: Somethign i've always wondered, do we care if the default values here are `undefined` or an empty string? if not we can probably use one of them.

mnatsheh: For the most part, I prefer undefined, as that should not be serialized JSON.

src/schema/nodes/media.ts
```diff
@@ -0,0 +1,55 @@
...
+export interface MediaNode extends ManuscriptNode {
+  attrs: Attrs
+}
+
+export const media: NodeSpec = {
+  attrs: {
+    id: { default: '' },
+    dataTracked: { default: null },
+    href: { default: '' },
+    show: { default: '' },
```
mnatsheh, line 38: We don't need to add this. When exporting, we should always use "embed".

mnatsheh (commented):
I feel that we should rename the type to "embed", since this is just for embedding.

src/schema/nodes/media.ts
mbetamony: Updated.

src/schema/nodes/media.ts
mbetamony: Updated.

mbetamony:
> I feel that we should rename the type to "embed", since this is just for embedding.

Renamed it.

mbetamony:
Please let me know if this is ready for QA, or if it still needs review.

mnatsheh (approved):

src/jats/importer/jats-transformations.ts
```diff
@@ -148,8 +148,11 @@ export const createBody = (
 ) => {
   const group = createSectionGroup(bodyType, createElement)
 
-  const paragraphs = doc.querySelectorAll('body > p')
-  group.append(...paragraphs)
+  const elements = body.querySelectorAll(':scope > *:not(sec)')
+  elements.forEach((element) => {
+    removeNodeFromParent(element)
+    group.appendChild(element)
+  })
```
mbetamony, line 155: Hi @mnatsheh, I've added support for all elements here instead of paragraphs. Is that okay? and if so, should we only include elements that are before the first section? 
I'm also working on a cleanup that would remove the need for this.

mnatsheh: That's fine. JATS doesn't allow anything (mostly) after `sec` so we don't need to worry about that.

mnatsheh: BTW, is this something we missed in the non-section support we added recently?

PR #165, opened 2024-11-24
LEAN-4146: LW-FY25-Q3-S1: Missing Placeholder for Non-Imported Table …

nghazzawi-atypon:
…and Figure Captions


src/jats/__tests__/__snapshots__/jats-importer.test.ts.snap
```diff
@@ -14051,17 +14096,18 @@ Object {
...
                           "attrs": Object {
                             "dataTracked": null,
                             "placeholder": "Caption...",
                           },
-                          "content": Array [
-                            Object {
-                              "text": "Radiologic response assessed by positron emission tomography computed tomography in patients with classical Hodgkin lymphoma treated by low-dose anti-PD1 antibodies pembrolizumab and nivolumab.",
-                              "type": "text",
-                            },
-                          ],
```
mbetamony, line 14064: Looking at the tests I don't think this will import data correctly. 
We probably need to do something like this inside jats-transformations.ts, and call it in parse-jats-article:
e.g:
```
parse-jats-article.ts
  addCaptionsToFigures(doc, createElement)
  moveTitle(front, createElement)
  ...
```
 ```
 jats-transformations.ts
export const addCaptionsToFigures = (
  doc: Document,
  createElement: CreateElement
) => {
  const figs = doc.querySelectorAll('fig')
  figs.forEach((fig) => {
    const caption = fig.querySelector('caption')
    if (!caption) {
      const caption = createElement('caption')
      const title = createElement('title')
      caption.appendChild(title)
      fig.appendChild(caption)
    }
    const title = fig.querySelector('caption > title')
    if (!title) {
      const title = createElement('title')
      fig.querySelector('caption')?.appendChild(title)
    }
  })
}
```
And something similar for tables.

File: src/schema/nodes/figcaption.ts [hunk 2]
```diff
@@ -23,7 +23,7 @@ export interface FigCaptionNode extends ManuscriptNode {
 }
 
 export const figcaption: NodeSpec = {
-  content: 'caption_title? caption*',
+  content: 'caption_title caption+',
```
mbetamony, line 26: We can probably keep those as is, I'm not sure if this will behave correctly on the editor.

mbetamony (approved):

PR #169, opened 2024-12-02
LEAN-4146: add placeholder captions to figures and tables

mbetamony:


mnatsheh:
Was there a problem with the solution proposed in https://github.com/Atypon-OpenSource/manuscripts-transform/pull/165 ? I'm not sure it's a good idea to modify the dom within the dom-parser.

mbetamony:
> Was there a problem with the solution proposed in #165 ? I'm not sure it's a good idea to modify the dom within the dom-parser.

I honestly just wanted to see what we can accomplish using the dom-parser, but you're right, its not the easiest to debug/maintain.

mbetamony:
Could someone review this please?

mnatsheh (approved):

PR #171, opened 2024-12-09
LEAN-4195: ignore some of the labels on import

mbetamony:
Removed labels for:
- figures
- box elements
- equations elements

I also added it as a rule to ignore labels in general.

Kept labels for:
- sections
- corresp


mnatsheh (approved):
Should we remove the label from the schema as well?

mbetamony:
> Should we remove the label from the schema as well?

I removed it from `table_element` and `box_element`(weirdly enough it wasn't in `figure_element`), if that's what you meant.

PR #175, opened 2024-12-19
fix migration

mbetamony:


src/schema/migration/migrate.ts
```diff
@@ -55,7 +55,7 @@ export function migrateFor(oldDoc: JSONNode, baseVersion: string) {
       continue
     }
     console.log('Migrating doc with script to version ' + script.toVersion)
-    migratedDoc = migrate(migratedDoc, script.migrateNode)
+    migratedDoc = migrate(migratedDoc, script.migrateNode.bind(script))
```
mnatsheh, line 58: I think you can just do this:
```
migratedDoc = migrate(migratedDoc, () => script.migrateNode())
```

mbetamony: Updated.

mnatsheh (approved):

mbartenev-atypon:
I wonder why it worked before... FYI you can use class method preserving 'this' by binding it in the constructor and avoiding redefining it per context i.e. this.someMethod = this.someMethod.bind(this)

mbartenev-atypon:
> I wonder why it worked before... FYI you can use class method preserving 'this' by binding it in the constructor and avoiding redefining it per context i.e. this.someMethod = this.someMethod.bind(this)

Found the answer to that - this is the first migration ever to refer to 'this' inside migrateNode method.

PR #184, opened 2025-02-10
LEAN-4310: add abstract types

mbetamony:


mnatsheh:
I think we should have a different section category group for graphical vs text abstract.

mnatsheh:
I forgot about import/export... we should implement it in this ticket if simple.

mbetamony:
> I forgot about import/export... we should implement it in this ticket if simple.

I've added support in this ticket. I'm importing the new abstracts as regular sections, and the key-image section as its own node. 
I've also added a new section category group for graphical abstracts, but not really sure why do we need this distinction?

mbetamony:
Could someone review this please?

mnatsheh (approved):

PR #188, opened 2025-02-11
LEAN-4267: LW-FY25-Q3-S6: Missing element type in the LH navigator fo…

nghazzawi-atypon:
…r box element


mbetamony (approved):

src/transformer/node-title.ts
```diff
@@ -82,6 +82,9 @@ export const nodeTitle = (node: ManuscriptNode) => {
     case nodes.image_element:
       return ''
 
+    case nodes.box_element:
+      return snippetOfNodeType(node, nodes.box_element)
+
```
mbetamony, line 87: We could probably return an empty string here, and it should work, similar to the nodes above.

nghazzawi-atypon: right, done

mbetamony: thanks

PR #193, opened 2025-02-18
LEAN-4309: update the export to handle multiple self-uris

mbetamony:


src/jats/exporter/jats-exporter.ts
```diff
@@ -555,26 +555,29 @@ export class JATSExporter {
       journalMeta.remove()
     }
 
-    const selfUri = findChildrenByType(
+    const selfUriAttachments = findChildrenByType(
       this.manuscriptNode,
       schema.nodes.attachment
-    ).find((result) => result.node.attrs.type === 'document')
+    )
+      .filter((result) => result.node.attrs.type === 'document')
```
mnatsheh, line 562: We don't need the filter in this case.

mnatsheh (approved):

src/jats/exporter/jats-exporter.ts
mbetamony: Yeah sure, removed.

PR #195, opened 2025-02-24
LEAN-4351: add support for running and short titles

mbetamony:


mnatsheh:
I wonder if it's better to add a "type" attr to the title node, and each title would have its own instance of the title node with a different type.

mbetamony:
> I wonder if it's better to add a "type" attr to the title node, and each title would have its own instance of the title node with a different type.

Looking at the [designs](https://www.figma.com/design/kzL2zJgnExGmJ2cyt9zm2x/Support-for-multiple-title-types?node-id=16904-11612&t=XaLqvHYyd5lf7tw6-0), they seem more of attributes to me than actual nodes. I also don't know which is better/easier for the FE to implement though. Let me know if you prefer we change them to multiple title nodes.

mnatsheh:
Interesting, I've not seen the designs. Although they do look like attributes, we're expecting TC to work on them like other body text, so I suspect we will have to treat them as separate nodes I think.

src/jats/exporter/jats-exporter.ts
```diff
@@ -422,13 +422,19 @@ export class JATSExporter {
...
     const titleGroup = this.document.createElement('title-group')
     articleMeta.appendChild(titleGroup)
-
     this.buildContributors(articleMeta)
-    if (titleNode) {
-      const element = this.document.createElement('article-title')
+
+    titleNodes.forEach((titleNode) => {
+      let element
+      if (titleNode.attrs.type === 'main') {
```
mnatsheh, line 429: Let's call this "primary".

mnatsheh (approved):

src/jats/exporter/jats-exporter.ts
mbetamony: Done

mnatsheh:
As discussed offline, it actually makes more sense to create a new node type for these alt titles.

src/jats/importer/jats-transformations.ts
```diff
@@ -57,15 +57,31 @@ export const addMissingCaptions = (
...
   let title = front.querySelector('article-meta > title-group > article-title')
   if (title) {
     title.innerHTML = htmlFromJatsNode(title, createElement) ?? defaultTitle
   } else {
     title = createElement('article-title')
     title.innerHTML = defaultTitle
   }
-  front.parentNode?.insertBefore(title, front)
+  titles.appendChild(title)
+  const runningTitle = front.querySelector(
```
mnatsheh, line 70: We can probably replace this with a loop.

mnatsheh (approved):

src/jats/importer/jats-transformations.ts
mbetamony: Sure, done.

PR #198, opened 2025-02-27
LEAN-4352: create a manuscripts validator base

mbetamony:
Might be an over kill for what we're building, but we might as well have it if we're going to have more validations in the future. The reason i didn't add it in the api as discussed yesterday was because I wanted to use `findChildrenByType`, and we need to add a PM utils dependency there to use it.


mnatsheh (commented):
It feels like we don't need any of these to be classes; can we replace them with functions?

mbetamony:
> It feels like we don't need any of these to be classes; can we replace them with functions?

Yes, done.

src/transformer/node-validator.ts
```diff
@@ -0,0 +1,89 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { Node, NodeType } from 'prosemirror-model'
+import { findChildrenByType } from 'prosemirror-utils'
+
+export interface ValidationError {
+  code: string
+  nodeType?: string
```
mnatsheh, line 21: I think we should change this to "id". This would allow us to find the exact problematic node.

src/transformer/node-validator.ts
```diff
@@ -0,0 +1,89 @@
...
+  isValid: boolean
+  errors: ValidationError[]
+}
+
+export const VALIDATION_ERROR_CODES = {
+  MISSING_TITLE: 'MISSING_TITLE',
+  EMPTY_TITLE: 'EMPTY_TITLE',
+}
+
+const validateTitleNode = (nodes: Node[]): ValidationResult => {
```
mnatsheh, line 34: I think we should just return the error list here. The "isValid" aggregation already happens at the top level.

mbetamony: I've updated the code, please let me know your thoughts.

src/transformer/node-validator.ts
```diff
@@ -0,0 +1,82 @@
...
+
+const validateNode = (
+  nodeType: NodeType,
+  node: Node,
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  doc: Node
+): ValidationError[] => {
+  switch (nodeType.name) {
+    case 'title':
+      return isTitleNode(node) ? validateTitleNode(node) : []
```
mnatsheh, line 54: Is the reason for the `isTitleNode` call to change the node type? Can't we just use `node as TitleNode`?

mnatsheh (approved):

src/transformer/node-validator.ts
mbetamony: Yeah, I initially did it to ensure type safety. But it might be simpler to just is `node as TitleNode`

PR #199, opened 2025-03-18
LEAN-4405: bibliography items work

mbetamony:


mnatsheh:
One thing I remembered is that the `BibliographyItemAttrs` structure was done to match what citeproc-js accepts (see https://github.com/Atypon-OpenSource/manuscripts-library/blob/417cd7ece753fe26660d23abc8372f3f1157f909/src/csl-converter.ts).

mbetamony:
> One thing I remembered is that the `BibliographyItemAttrs` structure was done to match what citeproc-js accepts (see https://github.com/Atypon-OpenSource/manuscripts-library/blob/417cd7ece753fe26660d23abc8372f3f1157f909/src/csl-converter.ts).


I've attempted to map the new fields to their corresponding fields in CSL-JSON. "x" means that i couldn't find a logical mapping.

- std → x
- data-title → x
- series → x
- publisher-name → publisher
- publisher-loc → publisher-place
- conf-name → event
- conf-date → event-date
- conf-loc → event-place
- institution → x
- ext-link → URL (Type: String)
- size @units-pages → number-of-pages
- date-in-citation → original-date
- pub-id → x

Here are a few questions though: 

1. We currently store DOI inside the bibliography item attributes, which has a mapping to DOI, and it is also included in pub-ids. Should we store it in both places?

2. For ext-link, we agreed to store both the type and content of ext-link. However, since CSL-JSON has URL as a string, how should we handle this? Do we need to map it differently?

3. Are the mappings above correct? For more details, check [csl-json package](https://www.npmjs.com/package/csl-json?activeTab=code) and check index.d.ts.

src/schema/nodes/bibliography_item.ts
```diff
@@ -24,16 +24,54 @@ export interface BibliographyItemAttrs {
...
+  data: 'dataset',
+  preprint: 'article-journal',
+}
+
+export const publicationTypeToJats: Record<string, string> = {
+  article: 'journal',
+  'article-journal': 'journal',
+  webpage: 'web',
+  dataset: 'data',
+}
```
mnatsheh, line 74: We should have these in the exporter.

src/jats/importer/jats-dom-parser.ts
```diff
@@ -204,115 +207,144 @@ export class JATSDOMParser {
...
-    }
-    const title = this.getHTMLContent(element, 'article-title')
-    if (title) {
-      attrs.title = title
-    }
-
-    const mixedCitation = element.querySelector('mixed-citation')
-
-    if (authorNodes.length <= 0) {
+    if (authorNodes.length === 0) {
```
mnatsheh, line 250: I don't think the is useful. Let's remove it.

src/jats/importer/jats-dom-parser.ts
```diff
@@ -204,115 +207,144 @@ export class JATSDOMParser {
...
+        getHTMLContent(element, 'article-title') ??
+        getHTMLContent(element, 'data-title') ??
+        getHTMLContent(element, 'part-title'),
     }
+    const fpage = getTrimmedTextContent(element, 'fpage')
+    const lpage = getTrimmedTextContent(element, 'lpage')
+    if (fpage) {
+      attrs.page = lpage ? `${fpage}-${lpage}` : fpage
+    }
+    this.parseRefPersonGroup(element, attrs)
```
mnatsheh, line 345: Can we replace this with something like the following? 
```
{
...
author: getNameContent(element, 'person-group[person-group-type="author"]'),
editor: getNameContent(element, 'person-group[person-group-type="editor"]'),
...
}
```

src/jats/importer/jats-dom-parser.ts
```diff
@@ -204,115 +207,144 @@ export class JATSDOMParser {
...
+    this.parseRefDates(element, attrs)
```
mnatsheh, line 346: Same as above.

src/jats/importer/jats-dom-parser.ts
mbetamony: That looks better, done.

src/jats/importer/jats-dom-parser.ts
mbetamony: Done

src/schema/nodes/bibliography_item.ts
mbetamony: Done

mbetamony:
A few notes: 
1- The export changes are not visible when you got to the PR changes, you'll need to expand them.
2- I forgot what we said about the conf-proc mapping to paper-conference, is that correct? I removed that mapping for now.

src/jats/importer/jats-dom-parser.ts
```diff
@@ -204,116 +206,127 @@ export class JATSDOMParser {
...
-    if (authorNodes.length <= 0) {
-      mixedCitation?.childNodes.forEach((item) => {
-        if (
-          item.nodeType === Node.TEXT_NODE &&
-          item.textContent?.match(/[A-Za-z]+/g)
-        ) {
-          attrs.literal = getTrimmedTextContent(mixedCitation) ?? ''
-          return attrs
-        }
+  private parseRefDates(element: Element, query: string) {
```
mnatsheh, line 240: Let's call this `getDateContent`.

src/jats/importer/jats-dom-parser.ts
```diff
@@ -204,116 +206,127 @@ export class JATSDOMParser {
...
     }
+  }
 
-    const supplement = getTrimmedTextContent(element, 'supplement')
-    if (supplement) {
-      attrs.supplement = supplement
+  private parseRefLiteral = (element: Element) => {
+    const mixedCitation = element.querySelector('mixed-citation')
+    for (const item of mixedCitation?.childNodes ?? []) {
+      if (
```
mnatsheh, line 262: Can we use `textContent` instead?

src/jats/importer/jats-dom-parser.ts
```diff
@@ -204,116 +206,127 @@ export class JATSDOMParser {
...
-          attrs.literal = getTrimmedTextContent(mixedCitation) ?? ''
-          return attrs
-        }
+  private parseRefDates(element: Element, query: string) {
+    const buildDate = (element: Element) => {
+      const isoDate = element.getAttribute('iso-8601-date')
+      if (!isoDate) {
+        return
+      }
+      const [datePart] = isoDate.split('T')
```
mnatsheh, line 246: Should we use `Date.parse`?

src/jats/importer/jats-dom-parser.ts
```diff
@@ -204,116 +206,127 @@ export class JATSDOMParser {
...
+        'person-group[person-group-type="author"] > *'
+      ),
+      editor: this.getNameContent(
+        element,
+        'person-group[person-group-type="editor"] > *'
+      ),
+      literal: this.parseRefLiteral(element),
+      accessed: this.parseRefDates(element, 'date-in-citation'),
+      'event-date': this.parseRefDates(element, 'conf-date'),
+      issued: getTrimmedTextContent(element, 'year')
```
mnatsheh, line 317: I think the query should be `> year`. otherwise we could pick up an incorrect date.

mnatsheh:
> I forgot what we said about the conf-proc mapping to paper-conference, is that correct? I removed that mapping for now.

It's fine. Let's use `confproc` instead of `conf-proc` as specified in the specs.

src/jats/importer/jats-dom-parser.ts
mbetamony: Renamed

src/jats/importer/jats-dom-parser.ts
mbetamony: Yeah that's simpler, done.

src/jats/importer/jats-dom-parser.ts
mbetamony: This will change the result we have. this requires direct _text_ nodes. I simplified the logic a bit though

src/jats/importer/jats-dom-parser.ts
mbetamony: Changed. I had to pass in the `element-citation` | `mixed-citation` to the trimmer.

mnatsheh:
We should map it to locator.

PR #203, opened 2025-03-19
LEAN-4425: support for alt-text and long-desc

mbetamony:
I opted to use attributes instead of having it as a node since not all the nodes containing it have content. Not sure if you prefer we have it as content anyway.


mnatsheh (approved):

mbetamony:
I remember we said we don't want to add these elements on the supplements, please let me know if i misunderstood.

src/jats/importer/jats-transformations.ts
```diff
@@ -454,6 +454,25 @@ export const createSupplementaryMaterialsSection = (
...
 }
 
+export const moveAccessibilityItems = (doc: Document) => {
+  const items = doc.querySelectorAll('alt-text, long-desc')
+  items.forEach((item) => {
+    item.parentNode?.appendChild(item)
+  })
+}
+
+export const createImages = (doc: Document, createElement: CreateElement) => {
```
mnatsheh, line 464: Why did we need to add this? was it not working before?

mbetamony: In the dom parser, i matched the 'graphic' tag with the 'image_element' node. And for that to keep working we'll need to return the content explicitly, since we'll need a figure node inside the graphic node, something like the newest commit

mnatsheh: but how did it work before? Matching graphic to image_element was already there before.

mbetamony: Similar to the newer implementation, but it was simpler since it was one piece of content. 
``` 
    getContent: (node) => {
        const figure = this.schema.nodes.figure.create(
          this.getFigureAttrs(node)
        )
        return Fragment.from(figure)
      },
```

mnatsheh: Got it.

mnatsheh (approved):

PR #208, opened 2025-03-25
LEAN-4406 Common parent for alt-titles

mbartenev-atypon:


src/schema/nodes/alt_titles_section.ts
```diff
@@ -0,0 +1,60 @@
+/*!
+ * © 2019 Atypon Systems LLC
+ *
```
mbetamony, line 3: should be 2025

src/schema/nodes/alt_titles_section.ts
```diff
@@ -0,0 +1,60 @@
...
+import { NodeSpec } from 'prosemirror-model'
+
+import { ManuscriptNode } from '../types'
+
+interface Attrs {
+  id: string
+}
+
+// TODO: endnotes section?
+
```
mbetamony, line 26: we can maybe remove this comment

src/schema/nodes/alt_titles_section.ts
```diff
@@ -0,0 +1,60 @@
...
+      },
+      0,
+    ]
+  },
+}
+
+export const isAltTitlesSectionNode = (
+  node: ManuscriptNode
+): node is AltTitlesSectionNode =>
+  node.type === node.type.schema.nodes.footnotes_section
```
mbetamony, line 60: wrong node type here

mbetamony (approved):

src/schema/nodes/alt_titles_section.ts
mbartenev-atypon: done

src/schema/nodes/alt_titles_section.ts
mbartenev-atypon: done, thanks!

mbetamony:
We'll probably need a migration for older docs to use the new node

mbartenev-atypon:
We discussed with Mohammad that given that alt-titles were introduced very recently, we probably don't need to worry about migrations for them. What do you think?

mbetamony:
> We discussed with Mohammad AlNatsheh that given that alt-titles were introduced very recently, we probably don't need to worry about migrations for them. What do you think?

that sounds ok to me

PR #215, opened 2025-04-25
LEAN-4432 Creating special node type for images inside quotes.

mbartenev-atypon:


mnatsheh (approved):

mnatsheh:
I'm assuming it was not possible to reuse the figure node implementation for the new node type.

mbartenev-atypon:
> I'm assuming it was not possible to reuse the figure node implementation for the new node type.

It wouldn't be possible to use the same one.  A couple of properties are different and it would've needed some sort of overriding/inheritance approach to reuse it. I think it is more straight forward just to create schema definitions without any relation to other nodeType schemas.

PR #223, opened 2025-05-08
LEAN-4529: add prefix and role to contributor

mbetamony:
In the previous implementation, we had role always set to `author`. We also export all contributor with the role author with a contrib-group of `author`, should i keep that?


mnatsheh (approved):
We don't need to worry about the "author" role. Note that the same <role> tag will be used for CRediT, so this handling might change a bit later.

PR #224, opened 2025-05-08
LEAN-4527: import/export labels for box_element

mbetamony:
I couldn't find any designs for the custom labels, so i reused the `section_label` node. Should we create a new node anyways? 
This is how this will look on the editor now: 

Screenshot 2025-05-08 at 7 34 17 PM


mnatsheh:
What we need to support is for the user to be able to _remove_ the figcaption that's already there. If the figcaption doesn't exist, we should not export a <label> for the box.

mnatsheh:
What about the label? I'm assuming we always export a label.

mbetamony:
> What about the label? I'm assuming we always export a label.

Yeah. Isn't what required here to have custom labels for box_elements?

mnatsheh:
No, what's needed is to completely remove the label and caption for these boxes. At least that's my understanding.

mbetamony:
> No, what's needed is to completely remove the label and caption for these boxes. At least that's my understanding.

I had a chat with Panos about what is exactly required here, and it turns out what we want is the following: 
1- keep the custom labels (e.g Box 1)
2- have the ability to toggle the label on/off for each box element

I've updated the box_element node to have a labelToggle attr, which we can add to other nodes if we decide to follow a similar approach. Let me know your thoughts, thank you.

mnatsheh:
Can't we depend on the existence of figcaption to show/hide the label as I suggested earlier?

mnatsheh:
I don't believe we should allow users to enter a caption for such boxes. Check also the screenshots in Confluence that include what the expectation is. Anyway, I'll discuss with Panos.

mbetamony:
> I don't believe we should allow users to enter a caption for such boxes. Check also the screenshots in Confluence that include what the expectation is. Anyway, I'll discuss with Panos.

Sure, please let me know.

mbetamony:
What i've done here is introduce a new attribute (captionToggle) on the box element. If the toggle is off, we don't export the caption and label, in other words, we don't care about the content. 
Not sure if we'll actually need a different node (e.g box_element_without_caption) for this to work.. let me know your thoughts, thanks.

mnatsheh:
I'm cycling back to my suggestion: Can't we depend on the existence of figcaption to show/hide the label as I suggested earlier?

mnatsheh (approved):

PR #226, opened 2025-05-14
LEAN-2023 - use pnpm

mnatsheh:
It seems like PM updated how DOMParser works, so I need to go through the tests and make sure everything is working correctly.


pnpm-workspace.yaml
```diff
@@ -0,0 +1,2 @@
+onlyBuiltDependencies:
+  - libxmljs2
```
asouqi, line 2: I think we need to add also
```
packages:
  - .
```

package.json
```diff
@@ -15,72 +15,72 @@
...
-    "prosemirror-state": "^1.4.2",
-    "prosemirror-view": "^1.29.1",
-    "rimraf": "^3.0.2",
-    "typescript": "^4.0.5"
+    "npm-run-all": "4.1.5",
+    "prettier": "2.8.8",
+    "prosemirror-state": "1.4.3",
+    "prosemirror-view": "1.39.2",
+    "rimraf": "6.0.1",
+    "typescript": "4.9.5"
```
mbetamony, line 84: It might be worth it to update typescript to a later version, there are some cool features like inferred type predicates(5.5) and decorators(5.0)

pnpm-workspace.yaml
mnatsheh: Can you explain more? it's working without it.

package.json
mnatsheh: I don't want to jump major versions in this PR. We can do it but separately.

mbartenev-atypon (approved):

PR #227, opened 2025-05-15
LEAN-4375 Preserving paragraphs & discarding unsupported footnotes in author-notes 

mbartenev-atypon:


src/jats/importer/jats-dom-parser.ts
```diff
@@ -501,6 +501,11 @@ export class JATSDOMParser {
         }
       },
     },
+    {
+      tag: 'fn[fn-type]', // all supported fn-types should be moved out by the time we parse into prosemirror
+      context: 'author_notes/',
+      ignore: true,
+    },
```
mbetamony, line 508: The remaining footnotes here are the ones that do not have a section category mapping right?

mbartenev-atypon: yes, if we didn't process some of the fn[fn-type] by this time, they are to be considered unsupported.

mbetamony (approved):

src/jats/importer/jats-dom-parser.ts
mbetamony: Got it, thanks!

PR #231, opened 2025-05-26
LEAN-4576: add unstructured bibliography item

mbetamony:
I've created a new node to handle this, please let me know if we should reuse the bibliography_item node instead.


mnatsheh:
I don't think we need a new node type. A new `BibliographyItemType` with only `literal` populated should be enough. I think most of the work will be in the FE.

mbetamony:
Got it, thanks. Updated.

mnatsheh (changes requested):
We need to add a bib item type. This will allow the app to display this correctly.

mnatsheh:
It's not incorrect per se, but we don't care about it. The example in the Literatum tagging guidelines doesn't have a type, so I think it's fine to ignore it even if it's there.

mbetamony:
> It's not incorrect per se, but we don't care about it. The example in the Literatum tagging guidelines doesn't have a type, so I think it's fine to ignore it even if it's there.

Got it, done. I reused the `getRefType` function from your [PR](https://github.com/Atypon-OpenSource/manuscripts-transform/pull/230).

mnatsheh (approved):

mbetamony:
I've changed the citeproc logic to exclude the publication type for unstructured references and add the specific-use attribute.

mnatsheh (approved):

PR #232, opened 2025-05-26
LEAN-4577: add support for trans-abstract

mbetamony:


src/jats/importer/parse-jats-article.ts
```diff
@@ -91,7 +91,7 @@ export const parseJATSArticle = (
 ) => {
   const journal = parseJournal(doc)
   processJATS(doc, sectionCategories)
-
+  const schema2 = schema
```
mnatsheh, line 94: I'm guessing this is from some testing?

mbetamony: Removed, thanks.

mbetamony:
I added some updates here, I'm still not sure if we need a title or not. Here's how it would like if we didn't hide it in the body-editor. 
Screenshot 2025-05-29 at 7 59 47 PM

src/jats/exporter/jats-exporter.ts
```diff
@@ -2021,9 +2037,13 @@ export class JATSExporter {
     abstractCategories: string[]
   ): Element[] {
     if (container) {
-      return Array.from(container.querySelectorAll(':scope > sec'))
+      return Array.from(
```
mnatsheh, line 2040: Is `container` ever `null`? Can we remove this function?

src/jats/exporter/jats-exporter.ts
```diff
@@ -1988,17 +1991,30 @@ export class JATSExporter {
...
 
-    if (abstractSections.length) {
-      this.processAbstractSections(abstractSections, front)
+    for (const abstractSection of abstractSections) {
+      const node =
+        abstractSection.nodeName === 'trans-abstract'
+          ? this.createTransAbstractNode(abstractSection)
+          : this.createAbstractNode(abstractSection)
+      abstractSection.remove()
+      const articleMeta = front.querySelector(':scope > article-meta')
```
mnatsheh, line 2007: This should not be needed.

src/jats/exporter/jats-exporter.ts
```diff
@@ -1988,17 +1991,30 @@ export class JATSExporter {
...
+      if (articleMeta) {
+        insertAbstractNode(articleMeta, node)
+      }
     }
 
+    const transAbstracts = body.querySelectorAll(':scope > trans-abstract')
+    transAbstracts.forEach((transAbstract) => {
+      insertAbstractNode(articleMeta, transAbstract)
+    })
```
mnatsheh, line 2016: This should not be needed; it should already be handled in the previous loop.

src/jats/exporter/jats-exporter.ts
```diff
@@ -2043,15 +2063,30 @@ export class JATSExporter {
     front: HTMLElement
   ) {
     for (const abstractSection of abstractSections) {
-      const abstractNode = this.createAbstractNode(abstractSection)
+      const node =
```
mnatsheh, line 2066: We should be able to remove the entire `processAbstractSections` function.

src/jats/importer/jats-transformations.ts
```diff
@@ -124,11 +124,13 @@ export const moveAffiliations = (
 }
 
 export const moveAbstracts = (
-  doc: Document,
+  front: Element,
   group: Element,
   createElement: CreateElement
 ) => {
-  const abstracts = doc.querySelectorAll('front > article-meta > abstract')
+  const abstracts = front.querySelectorAll(
```
mnatsheh, line 131: I think we should handle these separately. For trans-abstracts, we can use the node as is without any modifications.

src/jats/importer/jats-transformations.ts
```diff
@@ -293,11 +295,13 @@ const createAbstractSection = (
   createElement: CreateElement
 ) => {
   const abstractType = abstract.getAttribute('abstract-type')
-
-  const section = createElement('sec')
   const sectionType = abstractType ? `abstract-${abstractType}` : 'abstract'
+  if (abstract.nodeName === 'trans-abstract') {
```
mnatsheh, line 299: Based on the above comment, we can get rid of this.

mnatsheh (changes requested):

src/jats/exporter/jats-exporter.ts
mbetamony: Why isn't this needed? we'll need to query the article-meta to insert the abstract in its correct place

src/jats/exporter/jats-exporter.ts
mbetamony: surprisingly, container is always null (we're querying the wrong element it seems). I'll file a jats-exporter cleanup ticket after this to have a look at everything here and see if we can depend on the nodes instead of moving things around.

src/jats/exporter/jats-exporter.ts
mbetamony: Yes, removed. Thanks

src/jats/exporter/jats-exporter.ts
mnatsheh: article-meta is queried a few lines above.

PR #234, opened 2025-05-28
Lean 4533

fwaisi:


src/schema/nodes/affiliation.ts
```diff
@@ -43,6 +43,7 @@ export interface AffiliationNode extends ManuscriptNode {
 
 export const affiliation: NodeSpec = {
   content: 'inline*',
+  atom: true,
```
mbartenev-atypon, line 46: Why we need this marked as atom for LEAN04533?

fwaisi: Each will be compared separately.

mbartenev-atypon (approved):

PR #235, opened 2025-05-28
Import cleanup concept

mbetamony:
This is a draft I've been testing. Instead of using a general DOM parser, I opted for `parseDOM` for the nodes. However, there is a one major issue: ideally, we would call `DOMParser.fromSchema(schema)`, but we can't have a static schema that incorporates custom rules and depends on `sectionCategories`. I tried some other solutions, but it turns out that creating different schema objects—even if they contain the same values—results in them being treated as distinct. For example, you cannot call `findNodeByType(node, schema.nodes.backmatter)` if the node's schema and the schema you are using are not identical. I implemented a workaround, which you can review in [parse-jats-article.ts.](https://github.com/Atypon-OpenSource/manuscripts-transform/pull/235/files#diff-b4bda23062e5f585162837287f237c9b8647c0438d8a218438998d8f6350f3ddR102-R113) The reason I tried this approach is that if we are creating different parsers for various nodes, it may also be beneficial to include each node's rules within its own definition

This is still a work in progress and still not fully functional, there are many things to change (e.g move functions definitions somewhere appropriate). I'm not quiet sure if this is a better approach than what we have, but I wanted to know your thoughts before proceeding.


mnatsheh:
agreed. I think we can close this.

PR #237, opened 2025-06-02
LEAN-4573: import/export for clickable images

mbetamony:
I've opted to add the inline_graphic as an attribute instead of having to create a separate inline graphic node; since i don't  know how hard that will be to implement for the FE.


mbetamony:
Actually this will probably not work. Looking at the designs(I couldn't find links yet, but it was shared on the sprint planning meeting), should we introduce an inline_figure node? Is that difficult to display for the FE? If we reuse the image node, that means we can't have it inside a paragraph similar to the example we have.

mbetamony:
Just a few updates:

1- I'm moving all ext-link elements that have graphics inside them to their own separate paragraph. Please let me know if we shouldn't do it that way and keep them as is.

2- I didn't know if its better to create inline nodes for the alt-text and long-desc, let me know if we should convert them into their own nodes.

Thank you.

src/jats/exporter/jats-exporter.ts
```diff
@@ -690,7 +690,7 @@ export class JATSExporter {
           id: normalizeID(node.attrs.id),
         }
         if (node.attrs.lang) {
-          attrs['xml:lang'] = node.attrs.lang
+          attrs[`${XML_NAMESPACE} xml:lang`] = node.attrs.lang
```
mnatsheh, line 693: Can you elaborate on this change?

src/schema/nodes/image_element.ts
```diff
@@ -20,13 +20,15 @@ import { ManuscriptNode } from '../types'
 export interface ImageElementNode extends ManuscriptNode {
   attrs: {
     id: string
+    extLink: string
```
mnatsheh, line 23: Let's mark this as optional.

mnatsheh (approved):

src/jats/exporter/jats-exporter.ts
mbetamony: Yep, changed, thanks.

PR #238, opened 2025-06-02
LEAN-4608: support for subtitles

mbetamony:


src/jats/importer/jats-dom-parser.ts
```diff
@@ -410,8 +410,8 @@ export class JATSDOMParser {
       },
     },
     {
-      tag: 'titles',
-      node: 'alt_titles',
+      tag: 'subtitle',
```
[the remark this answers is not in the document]
mbetamony: Yes, we don't need to specify it.

src/schema/index.ts
```diff
@@ -151,12 +159,14 @@ export * from './nodes/link'
...
 export * from './nodes/listing'
 export * from './nodes/listing_element'
+export * from './nodes/long_desc'
 export * from './nodes/manuscript'
 export * from './nodes/missing_figure'
 export * from './nodes/paragraph'
 export * from './nodes/placeholder'
 export * from './nodes/placeholder_element'
 export * from './nodes/pullquote_element'
+export * from './nodes/quote_image'
```
mbartenev-atypon, line 169: isnt it duplicated? i see it's already exported on line 178/188

src/schema/index.ts
```diff
@@ -151,12 +159,14 @@ export * from './nodes/link'
 export * from './nodes/list'
 export * from './nodes/listing'
 export * from './nodes/listing_element'
+export * from './nodes/long_desc'
```
mbartenev-atypon, line 162: this appears to be duplicated too along with hero_image and other added export

src/schema/index.ts
mbetamony: yep, fixed, thanks!

src/schema/index.ts
mbetamony: fixed, thanks!

PR #243, opened 2025-06-22
LEAN-4642

fwaisi:


src/schema/types.ts
```diff
@@ -181,6 +181,7 @@ export type ManuscriptTemplate = {
   bundle: string
   title: string
   sectionCategories: SectionCategory[]
+  articleType?: string
```
mbetamony, line 184: Based on the description of the spec, i think this should be required.

src/jats/importer/jats-dom-parser.ts
```diff
@@ -370,7 +370,7 @@ export class JATSDOMParser {
         const dates = this.parseHistoryDates(history)
         return {
           doi: getTrimmedTextContent(doi),
-          articleType: element.getAttribute('article-type') ?? '',
+          articleType: element.getAttribute('article-type') || 'other',
```
mnatsheh, line 373: This is not correct. If no article-type is provided in the JATS, we should leave this as `undefined`. When exporting, the value of article-type should be the value in the manuscript node, or the value in the template if it's undefined in the manuscript node.

mnatsheh (changes requested):

PR #248, opened 2025-07-09
LEAN-4647: add a secondary node to represent external media

mbetamony:
We discussed this today, me and @fwaisi, that it might make more sense to introduce a second node to represent the internal media. The way we'd separate between the two is by checking the href id. I'll also update the WM to change the media's hrefs if they're not linking to external sources.


src/jats/importer/jats-dom-parser.ts
```diff
@@ -752,6 +752,23 @@ export class JATSDOMParser {
     {
       tag: 'media',
       node: 'embed',
+      getAttrs: (node) => {
+        const element = node as HTMLElement
+        const href = element.getAttributeNS(this.XLINK_NAMESPACE, 'href')
+        if (href && href.startsWith('attachment:')) {
```
mnatsheh, line 758: This is my biggest concern here; the transform should not know about the "attachment" scheme. This is why for me it makes more sense to use a single node, and have the FE switch the view based on whether a file exists for this href or not.

What were the concerns with using a single node?

fwaisi:
@fwaisi, I will delete this branch since most of the changes are not required anymore.

PR #249, opened 2025-07-14
LEAN-4647 UM-030 - Display/Manage/Insert inline media

fwaisi:


mbetamony:
Do we need the new attribute? Can't we depend on the href to figure out if its internal or external? The type will be undefined for imported nodes anyways.

fwaisi:
> > @mbetamony, this type attr for positioning, same as the figure.
> 
> Is it something we need to import/export?

@mbetamony, I am not sure. currently its used for UI positioning.

@mnatsheh, could you please advise?

mnatsheh:
We should not need to position for media. I think we can close this PR.

fwaisi:
the changes in this PR is no longer needed. closing and deleting the branch

PR #258, opened 2025-08-06
LEAN-4774: add captions to supplements 

mbetamony:
I added figcaption node to supplements, and also added alt_text and long_desc. Not sure if they need to be required, since designs are not ready yet.


mnatsheh (approved):
Let's wait for the designs to make sure we need to accessibility elements before merging.

PR #262, opened 2025-08-21
LEAN-4820: single-name authors

mbetamony:


src/jats/exporter/jats-exporter.ts
```diff
@@ -1389,8 +1389,23 @@ export class JATSExporter {
...
+            surname: family,
+            'given-names': given,
+            prefix: contributor.attrs.prefix,
+          }
+          Object.entries(nameMapping).forEach(([tag, value]) => {
+            if (value) {
+              this.appendElement(name, tag, value)
+            }
+          })
+          contrib.appendChild(name)
```
mnatsheh, line 1407: Not a big fan of this approach. It feels cleaner to just update the `buildContributorName` function.

src/jats/importer/jats-dom-parser.ts
```diff
@@ -552,7 +552,9 @@ export class JATSDOMParser {
             : undefined,
           bibliographicName: {
             given: getTrimmedTextContent(element, 'name > given-names'),
-            family: getTrimmedTextContent(element, 'name > surname'),
+            family:
+              getTrimmedTextContent(element, 'name > surname') ??
+              getTrimmedTextContent(element, 'string-name'),
```
mnatsheh, line 557: Not sure if it matters, but this will not behave correctly if a name only contains `<given-names>`.

mbetamony: Why not? It would return undefined, or am i missing something?

src/jats/exporter/jats-exporter.ts
mbetamony: By this approach, do you mean the object mapping? Or just re-introducing the function?

PR #274, opened 2025-10-01
Proper handling of empty list items (LEAN-4789)

asouqi:


src/jats/exporter/jats-exporter.ts
```diff
@@ -968,7 +969,8 @@ export class JATSExporter {
 
         return linkNode
       },
-      list_item: () => ['list-item', 0],
+      list_item: (node) =>
+        node.childCount === 0 ? ['list-item', ['p']] : ['list-item', 0],
```
mnatsheh, line 973: Since we're filling the items, do we need this?

PR #309, opened 2026-04-05
LEAN-5474: LW-FY26-Q1: The curser Inside "Pull Quote " is not display…

AlaAlkhalaileh:
…ing properly after adding an Image


src/schema/migration/migration-scripts/4.3.24.ts
```diff
@@ -0,0 +1,40 @@
...
+    if (node.type === 'pullquote_element') {
+      return {
+        ...node,
+        content: [
+          {
+            type: 'text_block',
+            attrs: {},
+            content: node.content,
+          },
+        ],
```
asouqi: we need to update just paragraphs node in the pullquote, so that should be:
```
        content: node.content?.map((child) =>
          child.type === 'paragraph' ? { ...child, type: 'text_block' } : child
        ),
```

asouqi (changes requested):

src/schema/migration/migration-scripts/4.3.24.ts
AlaAlkhalaileh: done, could you please check now, also please review https://github.com/Atypon-OpenSource/manuscripts-body-editor/pull/915

asouqi (approved):

PR #315, opened 2026-04-12
Prevent users from manipulating comments (LEAN-5385)

asouqi:


src/schema/nodes/comment.ts
```diff
@@ -44,6 +44,28 @@ export const comment: NodeSpec = {
     timestamp: { default: 0 },
     originalText: { default: '' },
   },
+  canInsertNode(_, context) {
```
mbartenev-atypon, line 47: I see that it's draft but I wonder if we really need to add permission methods on a node? The strategy approach you had in api looked fine to me.

asouqi: Added here to the node, so any change to the schema here will be maintained in the transform not API.

I did that here based on what I got from your comment:
> The other question is how deep we actually want to go with granularity control because if we will start check separate attributes any schema change will need a corresponding update here. This is not a problem but surely a time consuming commitment that will increase this project maintainability requirements. It would be great if we can keep it high level.

or do you think we need to check here access to just the attribute?

mbartenev-atypon: I was thinking we can just check if a user tries to edit any of attributes and deny changing them instead of checking if user can edit this or that attribute. With the exception of contents.

asouqi: hmm and what about other attribute like [resolve](https://github.com/Atypon-OpenSource/manuscripts-article-editor/blob/master/src/lib/capabilities.tsx#L60-L64)
```
    handleOwnComments: !isViewer,
    handleOthersComments: isOwner,
    resolveOwnComment: !isViewer,
    resolveOthersComment: isOwner || isEditor,
    createComment: !isViewer,
```

so we need to check with Panos that if the user has privilege to edit content that will be applied to all attributes will not be specifically for each attribute

mbartenev-atypon: What I want is for us to have a simple model to deal with. The question is how possible is it. What is your opinion on that?

asouqi: I think we need to keep node type definitions separate from the API, and it should be a generic enforcer that just asks "can this subject do this action?" without coupling to any specific node structure. This way when a new node type or attribute is added to the schema, the access rules are updated in the transform alone, and the API doesn't need any change.

So I get the inspiration from Attribute-Based Access Control model to define the access policies in the transforms, and API with a common interface `NodePolicy` will apply the policies without any maintenance to node types or attributes

mbartenev-atypon: ok, sounds good

mbartenev-atypon (approved):

src/lib/access-control.ts
```diff
@@ -0,0 +1,89 @@
...
+   *  - As an object per-attribute rules
+   *  - As a function one rule applied to all attribute changes*/
+  attrs?: Partial<Record<keyof T['attrs'], NodeRule<T>>> | NodeRule<T>
+}
+
+type NodesPolicy = {
+  [K in keyof ProtectedResources]?: NodePolicy<ProtectedResources[K]>
+}
+
+const nodesPolicy: NodesPolicy = {
```
mbartenev-atypon, line 66: Can you provide an example of how access policy should be created (in the code here) so a developer in future could easily do so for some of the nodes?

asouqi: updated

PR #316, opened 2026-04-14
ALO-010 - lock contributor from editing with imported ORCID (LEAN-5532)

asouqi:


mbetamony:
I'm not sure, but i remember in one of the planning calls we said we're going to identify the external ORCID using a specific attribute.

asouqi:
> I'm not sure, but i remember in one of the planning calls we said we're going to identify the external ORCID using a specific attribute.

@mbetamony as far as you remember is it [authenticated](https://jats.nlm.nih.gov/archiving/tag-library/1.1/attribute/authenticated.html)?

mnatsheh:
@asouqi yes, it's @authenticated.

asouqi:
> @asouqi yes, it's @authenticated.

update to a new attribute `isAuthenticated` and will export if it's true

src/jats/importer/jats-dom-parser.ts
```diff
@@ -564,6 +564,10 @@ export class JATSDOMParser {
             element,
             'contrib-id[contrib-id-type="orcid"]'
           ),
+          isAuthenticated:
+            element
+              .querySelector('contrib-id')
```
mnatsheh: We need to check the orcid specifically, not any id.

asouqi: updated

mnatsheh (approved):

PR #324, opened 2026-04-29
Headshot Grid (LEAN-5609)

asouqi:
@mnatsheh can you please check if the structure of the nodes looks correct, as I want to make sure I get it right before starting with the implementation.

Based on what I get from the sprint planning call, the grid will be at the end of body. but Panos [says](https://confluence.prod.atypon.com/wiki/spaces/E2E/pages/389447685/Headshot+grid) we can insert that grid anywhere in the body like a table, figure! 
>The Headshot Grid is a new inline element that can be inserted anywhere within the main body of a document


mnatsheh:
You should confirm on the ticket what's expected. 

The "headshotsGrid" node should be called "headshotGrid". If we want it to be insertable anywhere, we should add it to the _element_ group. Otherwise, we should add it to the sections group. 

Either way, the "headshot" node should be an element called "headshot_element" to match other similar items.

mnatsheh (commented):
I believe there is a mistake in the requirements. The JATS should _not_ include a `<fig>`.

asouqi:
> I believe there is a mistake in the requirements. The JATS should _not_ include a `<fig>`.

updated

src/jats/importer/jats-dom-parser.ts
```diff
@@ -788,6 +788,48 @@ export class JATSDOMParser {
...
+          wrapper.appendChild(altText)
+        }
+
+        const longDesc = element.querySelector('long-desc')
+        if (longDesc) wrapper.appendChild(longDesc)
+
+        return this.parse(wrapper, {
+          topNode: this.schema.nodes.headshot_element.create(),
+        }).content
+      }
```
mbetamony, line 831: You can use `addMissingCaptions` from `jats-transformations.ts`, would that not work?

src/schema/nodes/headshot_image.ts
```diff
@@ -0,0 +1,63 @@
...
+export interface HeadshotImageAttrs {
+  id: string
+  src: string
+}
+
+export interface HeadshotImageNode extends ManuscriptNode {
+  attrs: HeadshotImageAttrs
+}
+
+export const headshotImage: NodeSpec = {
```
mbetamony, line 31: Do we need this node? Can't we use figure directly?

src/jats/importer/jats-dom-parser.ts
asouqi: Hmm I try it, but that did not work for me to move out caption

src/schema/nodes/headshot_image.ts
asouqi: In the [design](https://www.figma.com/design/siUx5nrIhiQEsk651JEyI2/Contributor-Headshot-Grid?node-id=41915-16493&m=dev) it shows just an image. and if we use figure will end up with a lot of checks in the body-editor for the [figure node view](https://github.com/Atypon-OpenSource/manuscripts-body-editor/blob/master/src/views/figure_element.ts#L27) to not show buttons around image for the headhost

Screenshot from 2026-05-14 17-31-56


Screenshot from 2026-05-14 17-34-18

mnatsheh: Yes, let's keep the separate node type similar to quote_image. I see that node type has a "type" attr, but I'm not sure what it's used for. Let's check what it is and maybe add it here as well.

src/schema/nodes/headshot_element.ts
```diff
@@ -0,0 +1,48 @@
...
+
+import { NodeSpec } from 'prosemirror-model'
+
+import { ManuscriptNode } from '../types'
+
+export interface HeadshotElementNode extends ManuscriptNode {
+  attrs: {
+    id: string
+    priority: number
+  }
```
mnatsheh, line 25: Let's have a separate Attrs type.

src/schema/nodes/headshot_element.ts
```diff
@@ -0,0 +1,48 @@
...
+ */
+
+import { NodeSpec } from 'prosemirror-model'
+
+import { ManuscriptNode } from '../types'
+
+export interface HeadshotElementNode extends ManuscriptNode {
+  attrs: {
+    id: string
+    priority: number
```
mnatsheh: Do we still use priority for order? @mbartenev-atypon

src/schema/nodes/headshot_image.ts
asouqi: In the editor we use it to set image alignment ['half-left', 'half-right'], but without affecting the actual CSS of the figure. and we exported in graphic as `fig-type` attribute

src/schema/nodes/headshot_element.ts
asouqi: done

src/schema/nodes/headshot_element.ts
asouqi: we use it for [Contributor](https://github.com/Atypon-OpenSource/manuscripts-body-editor/blob/522cb4b955d0f2c3370723454a489d30190e05ca/src/components/authors/AuthorsModal.tsx#L275)

mbartenev-atypon: @asouqi do you think we can rely on the nodes order in the document for order? Authors do still rely on priority but they don't have any inline text that needs to be edited right in the body and therefore it is easy to delegate their rendering to their parent element that also relies on priorities to sort and rerender them. head_element has a lot of inline editable content so it will be very hard to make it work in the same way as authors do.

asouqi: ooh that makes sense. What about relying on the track changes move operation, as we did for the [supplement node view](https://github.com/Atypon-OpenSource/manuscripts-body-editor/blob/522cb4b955d0f2c3370723454a489d30190e05ca/src/views/supplement.ts)

mbartenev-atypon: I think that will work well. This implies relying on natural order of elements in the document.

asouqi: removed that attribute

src/schema/nodes/headshot_element.ts
mnatsheh: I still don't see it.

mnatsheh (approved):

src/schema/nodes/headshot_element.ts
asouqi: I added It's on headshot_image, should I move it to the element?

mnatsheh: Each node should have its own Attrs type.

asouqi: updated, we should map it with JATS like this:
headshot_element: `content-type`
headshot_image:    `fig-type`

PR #342, opened 2026-06-02
LEAN-5625: add content-type to various elements

mbetamony:


mnatsheh:
I haven't checked the full PR yet, but I few points I noticed:
- I believe the attr in other node types is called "type". 
- We should not have any dependency on the value of this attr. i.e. we should not have "half-left" and similar in the source. We should be able to handle any value of the attr.

mbetamony:
> I haven't checked the full PR yet, but I few points I noticed:
> 
> * I believe the attr in other node types is called "type".
> * We should not have any dependency on the value of this attr. i.e. we should not have "half-left" and similar in the source. We should be able to handle any value of the attr.

Sure, updated.

mnatsheh (approved):

src/jats/__tests__/__fixtures__/jats-import.xml
```diff
@@ -251,6 +251,21 @@ xlink:href="prodbe.int.aip.org:8080/journal/journal_fs/
...
+    <disp-quote content-type="half-left">
+      <p>A half-left pullquote.</p>
+    </disp-quote>
+    <disp-quote content-type="half-right">
+      <p>A half-right pullquote.</p>
+    </disp-quote>
+    <disp-quote>
+      <p>pullquote without content-type</p>
+    </disp-quote>
+    <disp-quote content-type="quote">
```
mbartenev-atypon, line 283: @mbetamony so pull quote can never have alignment instructions, right?

mbetamony: Do you mean the blockquote? The examples above are pull quotes with alignments

mbartenev-atypon: Ah, yes. Sorry for the confusion. Indeed, I wonder if we need to have a capability to provide alignment to a blockquote.

mbetamony: We can add it, if that's desired. Please let me know if you need it.

mbartenev-atypon: No, not really. I was wondering what is the extent of support for content type. All is good from my side.

PR #344, opened 2026-06-11
LEAN-5686

mbetamony:


src/jats/exporter/jats-exporter.ts
```diff
@@ -313,7 +315,13 @@ export class JATSExporter {
     )
     engine.setOutputFormat('jats')
 
-    const output = engine.rebuildProcessorState([...citations.values()])
+    const uncitedItemIDs = includeUncitedReferences ? [...bibitems.keys()] : []
```
mnatsheh: how does adding all bibitem keys as uncited work?

mbetamony: It does the filtering internally. But, I agree that it's better to filter it for clarity.

mbetamony:
My understanding is that this is the only PR needed for now. If I've missed anything, please let me know. Thanks!

mnatsheh (approved):

PR #346, opened 2026-06-11
lean-prod - Opening previously created Drafts results in "Document not found" error in Editor for some items (LEAN-5747)

asouqi:


mnatsheh (approved):

asouqi:
The update in the migration test caught that issue. @mnatsheh but I want double check from you if we need to test that with QA and if yes should I set that up by deploying an old version of the app to mimic that article on the production?

mnatsheh:
@asouqi if you can verify that yourself using the steps you mention, we don't need to involve QA. First we need to deploy master to make sure the document _doesn't_ open then deploy this version and check that it does.

PR #351, opened 2026-06-29
LEAN-5683

fwaisi:


src/jats/exporter/labels.ts
```diff
@@ -41,8 +42,11 @@ const labelledNodeTypes: ManuscriptNodeType[] = [
   schema.nodes.box_element,
   schema.nodes.embed,
   schema.nodes.image_element,
+  schema.nodes.supplement,
 ]
 
+const noLabelTypes = new Set([schema.nodes.supplement.name])
```
mbartenev-atypon, line 48: I think this is duplicated with body-editor's type that excluded it for labels too. It would be best to use it in the body-editor too to avoid duplication

fwaisi: They actually cover different use cases. This one excludes the numeric label count for supplement nodes, while the one in the body editor excludes supplement nodes from decoration creation.

mbartenev-atypon: hm, isnt that decoration related to numeric labeling?

mbartenev-atypon: in body editor you omit labels by skpiing decorations creation for supplements, whether it is numbered or not. Here is just another case of the same thing really - you omit lables for supplements.

fwaisi: Yes, The intent was to keep label generation and decoration logic separate, so we can support cases where a node needs a label but no decoration, and vice versa. For now, I removed the body-editor exclusion logic, and decoration is skipped for any node that has no label.

mbartenev-atypon (approved):

PR #352, opened 2026-07-05
LEAN-5287: Monorepo split

mbetamony:
Hello,

This PR restructures the repository into a monorepo that publishes two independently versioned packages: `@manuscripts/schema` and `@manuscripts/transform`.

The main motivations for this change are:

1. **Consumer isolation**: Most repositories only depend on the schema package and do not require the transform package. `manuscripts-api` is the only repository that still depends on `@manuscripts/transform`.
2. **Independent versioning**: Changes to the transform package will no longer require version updates for repositories that only consume the schema package.

Please let me know your thought and concerns, and if you'd prefer to keep things as is. 

The copilot overview below seems to be accurate in terms of what changed. I'll also update the publishing pipeline before merging this, if we decide to go with it.


mbartenev-atypon:
I'm not really sure how useful this could be. About independent versioning: cases when we make a change in schema and it doesn't require a change in export/import almost never happen. So both repos will require corresponding changes for almost each feature or fix and versioning them independently will only add more work to submit PRs, publish etc. Also body editor shares some utilities with transform such as buildTargets and will have 2 dependencies instead of the single transform. This may be the case for article-editor too.
Additionally it will introduce an opportunity for a version mismatch during merge where schema and transform may be outdated in some of the dependencies and resolved incorrectly in the build (caused by pipeline transient hiccup for example). Considering that transform of version X could never work with any other version of schema besides corresponding version X - it will create a chance for a buggy release. In other words splitting this monolith will just make it a hidden monolith and not a dependency/consumer relation.
But I think it's a great idea to have some sort of backend package for isomorphic stuff - recently we've added capablitites into transform just because we need it both in backend and frontend and there simply no other place to put it.

mnatsheh:
If we were to split the repo, I think extracting the JATS code into its own package is the better way of doing it (as opposed to schema). @mbartenev-atypon thoughts?

mbartenev-atypon:
> If we were to split the repo, I think extracting the JATS code into its own package is the better way of doing it (as opposed to schema). @mbartenev-atypon thoughts?

I'm not really convinced that this is beneficial for us. This will allow the FE to avoid including the JATS code into the frontend bundles and that is the main difference that we will see as a result of decoupling the JATS export/import. But it is most likely that it is tree-shaked anyway and if it is not we should make sure that it is. So to achieve just that, it seems, that dealing with monorepo is a bit of an overkill. Additionally things like buildTargets which are shared among body-editor and JATS exporters will have to be shared somehow anyway, so relying on tree shaking maybe an easier option too for that. Otherwise I don't really mind this change.

mbetamony:
> Also body editor shares some utilities with transform such as `buildTargets` and will have 2 dependencies instead of the single transform. This may be the case for article-editor too.

I've actually moved all of the shared functionality into `schema`, so only the API will depend on `transform`.

> cases when we make a change in schema and it doesn't require a change in export/import almost never happen

I actually see the opposite happening more often. There are occasionally changes to the import/export logic that don't require any schema changes. That said, those cases are fairly rare as well.

> If we were to split the repo, I think extracting the JATS code into its own package is the better way of doing it (as opposed to schema)

Could you elaborate on why you think extracting the JATS code into its own package would be the better approach?

The main reason I suggested a monorepo with separate `schema` and `transform` packages is that the transformation code is tightly coupled to the schema. Keeping them in the same repository makes development much smoother, since implementing new import/export features often requires changes to both packages. Working in a single repo avoids the overhead of coordinating changes across multiple repositories while still allowing consumers that only need the schema to avoid depending on the transformation code.

mbartenev-atypon:
> The main reason I suggested a monorepo with separate `schema` and `transform` packages is that the transformation code is tightly coupled to the schema. Keeping them in the same repository makes development much smoother, since implementing new import/export features often requires changes to both packages. Working in a single repo avoids the overhead of coordinating changes across multiple repositories while still allowing consumers that only need the schema to avoid depending on the transformation code.

I agree about keeping them in the single repo, however splitting them into different packages in monorepo doesn't change much in terms of consumers of the dependencies. We can just as well keep it as a single package - FE consumers don't really depend on transformation code - it just sits there without making any difference. The problem of not including idle code into the build is a build related problem - such code should be tree-shaked and not moved out. Additionally, JATS code is a small dependency and even if it weren't tree-shaked it wouldn't affect the bundle in any significant way. 

Also I can get it when we put things like capabilities into transform because it's the only isomorphic repo we keep but when we create a completely new package - that is, "schema" in this case and immediately put there things that don't belong to schema by any measure: buildTargets and capabilities - this just tells me we need a better organisation of our code and dividing it like that doesn't solve much.

One other thing I realized is that article editor checks its transform version against api's transform and issues a warning if there is a mismatch. So this should be either disabled or adjusted somehow. It was added to warn developers on why, for example, the steps aren't accepted on BE.

mbetamony:
> > The main reason I suggested a monorepo with separate `schema` and `transform` packages is that the transformation code is tightly coupled to the schema. Keeping them in the same repository makes development much smoother, since implementing new import/export features often requires changes to both packages. Working in a single repo avoids the overhead of coordinating changes across multiple repositories while still allowing consumers that only need the schema to avoid depending on the transformation code.
> 
> I agree about keeping them in the single repo, however splitting them into different packages in monorepo doesn't change much in terms of consumers of the dependencies. We can just as well keep it as a single package - FE consumers don't really depend on transformation code - it just sits there without making any difference. The problem of not including idle code into the build is a build related problem - such code should be tree-shaked and not moved out. Additionally, JATS code is a small dependency and even if it weren't tree-shaked it wouldn't affect the bundle in any significant way.
> 
> Also I can get it when we put things like capabilities into transform because it's the only isomorphic repo we keep but when we create a completely new package - that is, "schema" in this case and immediately put there things that don't belong to schema by any measure: buildTargets and capabilities - this just tells me we need a better organisation of our code and dividing it like that doesn't solve much.

Yeah I agree... Thanks for the explanation, and I didn't actually know about tree shaking before, so definitely learnt something!

mnatsheh:
If there is an agreement that we don't need this, let's close the PR.

PR #353, opened 2026-07-05
LEAN-5777: Fix abstracts ordering

mbetamony:


mnatsheh:
@mbetamony how about creating a dedicated abstract node? would that be problematic?

src/schema/nodes/abstracts.ts
```diff
@@ -20,7 +20,7 @@ import { NodeSpec } from 'prosemirror-model'
 // It exists for the purpose of styling in the UI
 
 export const abstracts: NodeSpec = {
-  content: 'sections* (trans_abstract | trans_graphical_abstract)*',
+  content:'(section | graphical_abstract_section)* (trans_abstract | trans_graphical_abstract)*',
```
mnatsheh: I feel that we should have the order be even stricter: `section* graphical_abstract_section* trans_abstract* trans_graphical_abstract*`.

mbetamony:
> @mbetamony how about creating a dedicated abstract node? would that be problematic?

No, we can do that. To finalize implementation, would you prefer the abstracts keep their group as `group: 'block sections'` or have a new group for abstract sections, e.g `group: 'block abstract'`

mnatsheh:
@mbartenev-atypon any thoughts on adding a dedicated absatrct node?

> No, we can do that. To finalize implementation, would you prefer the abstracts keep their group as `group: 'block sections'` or have a new group for abstract sections, e.g `group: 'block abstract'`

I wonder if we even need the "sections" category.

mbartenev-atypon:
> @mbartenev-atypon any thoughts on adding a dedicated absatrct node?
> 
> > No, we can do that. To finalize implementation, would you prefer the abstracts keep their group as `group: 'block sections'` or have a new group for abstract sections, e.g `group: 'block abstract'`
> 
> I wonder if we even need the "sections" category.

Yes, sounds good. Although section category seems useful as we do indeed have sections group of elements.

mbetamony:
Hi again, I've updated the code. 

1. I introduced an abstract node, and kept its group `block sections`.
2. I did some cleanup to how we export abstracts, we used to serialize it inside the body, and move it around. 
3. I removed the 'abstract-' prefix from the abstract categories, we used to add it in import and remove it on export. (This will require an update to leanworkflow-enviornments, but I think it's worth it)
4. I've also moved most section categories code into the section-categories file to keep things centralized there.

src/jats/importer/jats-transformations.ts
```diff
@@ -156,18 +176,23 @@ export const moveAbstracts = (
...
-    group.appendChild(sec)
-  })
+
+  sections
+    .map((sec, index) => ({ sec, index }))
+    .sort((a, b) => {
+      const orderA = ABSTRACT_NODE_ORDER.indexOf(getAbstractNodeType(a.sec))
+      const orderB = ABSTRACT_NODE_ORDER.indexOf(getAbstractNodeType(b.sec))
+      return orderA - orderB || a.index - b.index
+    })
```
asouqi: @mbetamony are we sure that sort to nodes is necessary? based on the schema update on the parse to PM node `prosemirror-transform` it's going to place them in the right order

asouqi:
@mbetamony why did we return back `sections` to the groups of these nodes `graphical_abstract_section, trans_abstract, trans_graphical_abstract`

src/jats/exporter/jats-exporter.ts
```diff
@@ -229,12 +231,13 @@ export class JATSExporter {
     this.footnoteLabels = generateFootnoteLabels(manuscriptNode)
     const body = this.buildBody()
     article.appendChild(body)
+    this.buildAbstracts(front)
```
mnatsheh: should this be under `buildFront`?

src/jats/exporter/jats-exporter.ts
```diff
@@ -229,12 +231,13 @@ export class JATSExporter {
...
+    this.fixBody(front)
```
mnatsheh: Why do we need to call `FixBody` here?

src/jats/importer/jats-dom-parser.ts
```diff
@@ -1112,6 +1094,19 @@ export class JATSDOMParser {
...
       },
     },
+    {
+      tag: 'abstract',
+      node: 'abstract',
+      getAttrs: (node) => {
+        const element = node as HTMLElement
+        return {
+          id: element.getAttribute('id'),
+          category: element.getAttribute('abstract-type')
```
mnatsheh: Do we need the default to be "abstract"? why not keep it empty?

src/lib/section-categories.ts
```diff
@@ -25,3 +25,38 @@ export const getGroupCategories = (
...
+      }
+    }
+    return false
+  })
+}
+
+export const abstractTypeToCategory = (abstractType: string | null) =>
+  abstractType || 'abstract'
+
+export const FOOTNOTE_SECTION_CATEGORY_IDS = [
```
mnatsheh, line 51: This list doesn't match the one removed. Is it intentional?

mbetamony:
Hi again, A few additional changes I made:

* I also centralized the body/backmatter export routing.
* I replaced the scattered post-serialization DOM passes (where `moveNestedBodySectionsToBack` re-walked the already-built DOM and duplicated the same category logic used for top-level nodes) with a single `computeBackRouting` pass. Every section, regardless of nesting depth, is classified once and routed before serialization, instead of first being serialized into `<body>` and then queried back out afterward.
* I also finished applying the `$`-prefix convention to the remaining local DOM element variables in the serializer for consistency with the rest of the file.

I think this makes the pipeline more straightforward and easier to follow than what's currently on `master`. Rather than serializing these nodes into the body and moving them afterward, they're routed correctly from the start.

I still think there's more cleanup that could be done in this area. I made a few additional improvements while I was working on it, but I left them out because I didn't want this PR to grow any further.

If you don't agree with this approach, or if you'd rather keep these changes in a separate ticket/PR, I'm happy to revert them and split them out.

The main reason I included this here is that I had already started down this path while working on `buildFront`. As I dug deeper, I found the existing flow was more complex than I initially expected, and it felt cleaner to address the routing at the same time rather than continue building on top of it.

mbetamony:
I'll also have another review tomorrow and check if everything is correct, just wanted to add a headsup.

src/jats/exporter/jats-exporter.ts
mbetamony: yes, moved.

src/jats/exporter/jats-exporter.ts
mbetamony: We didn't, and i actually removed the function entirely.

src/jats/importer/jats-dom-parser.ts
mbetamony: I matched how it was on master to be honest. I removed it.

src/lib/section-categories.ts
mbetamony: No, thanks for noticing this. readded

mbetamony:
> @mbetamony why did we return back `sections` to the groups of these nodes `graphical_abstract_section, trans_abstract, trans_graphical_abstract`

I took a second look and body-editor uses `isSectionNodeType` which checks the `sections` group. So it either meant the isSectionNodeType will include new groups or we keep it as is, since we do want them to behave like sections If i'm not mistaken. Also based on @mbartenev-atypon comment above that it might be useful to keep them.

src/jats/importer/jats-transformations.ts
mbetamony: I don't think it does, i tried that and the tests failed. I think it walks DOM children in document order and tries to fit each one into the content-match state as it currently stands. So if we receive them in a wrong order, some nodes will be dropped.

asouqi:
> > @mbartenev-atypon any thoughts on adding a dedicated absatrct node?
> > > No, we can do that. To finalize implementation, would you prefer the abstracts keep their group as `group: 'block sections'` or have a new group for abstract sections, e.g `group: 'block abstract'`
> > 
> > 
> > I wonder if we even need the "sections" category.
> 
> Yes, sounds good. Although section category seems useful as we do indeed have sections group of elements.

@mbartenev-atypon should they be in a new group? As keeping them in a sections group means that the schema will allow inserting them at a section not just in an `abstracts` node, if I'm not mistaken

src/jats/exporter/jats-exporter.ts
```diff
@@ -338,26 +329,26 @@ export class JATSExporter {
       return null
     }
 
-    const template = this.createElement('template')
+    const $template = this.createElement('template')
 
-    template.innerHTML = JATSFragment
+    $template.innerHTML = JATSFragment
```
mnatsheh: we should rename `JATSFragment` to `fragment` or similar.

src/jats/exporter/jats-exporter.ts
```diff
@@ -382,192 +373,148 @@ export class JATSExporter {
...
-      this.appendCaption(supplementaryMaterial, node)
+  private buildSupplements = (): Node[] =>
+    this.getChildrenOfType(schema.nodes.supplement).map((node) =>
+      this.serializeNode(node)
+    )
 
-      articleMeta.append(supplementaryMaterial)
+  private buildHistory = (): Element | undefined => {
+    const $history = this.createElement('history')
+    const dates: Array<[number | undefined, string]> = [
```
mnatsheh: Should we use a map here? feels weird to use an array.

src/jats/exporter/jats-exporter.ts
```diff
@@ -382,192 +373,148 @@ export class JATSExporter {
...
-    ).length
-    countingElements.push(
-      this.buildCountingElement('ref-count', referencesCount)
-    )
+  }
 
-    //todo: is this correct?
-    const wordCount = this.manuscriptNode.textContent.split(/\s+/).length
-    countingElements.push(this.buildCountingElement('word-count', wordCount))
+  private buildCounts = (): Element | undefined => {
```
mnatsheh: Let's handle this the same way we handle other elements; add directly to the node then check if the node has any children.

src/jats/exporter/jats-exporter.ts
```diff
@@ -382,192 +373,148 @@ export class JATSExporter {
...
   }
 
+  private buildSelfUris = (): Node[] =>
+    this.getChildrenOfType(schema.nodes.attachment).map((attachment) =>
+      this.serializeNode(attachment)
+    )
+
   protected buildDateElement = (timestamp: number, type: string) => {
-    const dateElement = this.createElement('date')
+    const $dateElement = this.createElement('date')
```
mnatsheh: We can call this `$date`.

src/jats/exporter/jats-exporter.ts
```diff
@@ -577,80 +524,224 @@ export class JATSExporter {
...
   }
   protected buildCountingElement = (
     tagName: string,
     count: number | undefined
   ) => {
     if (count) {
-      const wordCount = this.createElement(tagName)
-      wordCount.setAttribute('count', String(count))
-      return wordCount
+      const $countElement = this.createElement(tagName)
```
mnatsheh: We can call this `$count`.

src/jats/exporter/jats-exporter.ts
```diff
@@ -577,80 +524,224 @@ export class JATSExporter {
...
+        node.attrs.category === 'body'
+      ) {
+        node.forEach((child) => {
+          if (child.type === schema.nodes.section) {
+            bodyNodes.push(child)
+          }
+        })
+        return
+      }
+      bodyNodes.push(node)
```
mnatsheh: what's the value of this?

src/jats/exporter/jats-exporter.ts
```diff
@@ -679,111 +770,100 @@ export class JATSExporter {
...
       embed: (node) => {
         const { id, href, mimetype, mimeSubtype } = node.attrs
         if (!href) {
           return ''
         }
-        const mediaElement = this.createElement('media')
-        mediaElement.setAttribute('id', normalizeID(id))
-        mediaElement.setAttributeNS(XLINK_NAMESPACE, 'show', 'embed')
-        mediaElement.setAttributeNS(XLINK_NAMESPACE, 'href', href)
+        const $mediaElement = this.createElement('media')
```
mnatsheh: Can we called `$media`.

src/jats/exporter/jats-exporter.ts
```diff
@@ -679,111 +770,100 @@ export class JATSExporter {
...
+        this.appendChildNodeOfType($mediaElement, node, schema.nodes.long_desc)
+        this.appendCaption($mediaElement, node)
+        return $mediaElement
       },
       awards: () => ['funding-group', 0],
       award: (node) => {
         const awardGroup = node as AwardNode
-        const awardGroupElement = this.createElement('award-group')
-        awardGroupElement.setAttribute('id', normalizeID(awardGroup.attrs.id))
+        const $awardGroupElement = this.createElement('award-group')
```
mnatsheh: Can be called `$awardGroup`.

src/jats/exporter/jats-exporter.ts
```diff
@@ -854,43 +962,44 @@ export class JATSExporter {
...
+          $fragment.appendChild(this.serializeNode(child))
+        })
+        return $fragment
       },
       inline_equation: (node) => {
         if (!node.attrs.contents) {
           return ''
         }
-        const eqElement = this.createElement('inline-formula')
+        const $equationElement = this.createElement('inline-formula')
```
mnatsheh: Maybe rename to `$inlineFormula`.

src/jats/exporter/jats-exporter.ts
```diff
@@ -944,38 +1056,38 @@ export class JATSExporter {
           return text
         }
 
-        const linkNode = this.createElement('ext-link')
-        linkNode.setAttribute('ext-link-type', 'uri')
-        linkNode.setAttributeNS(XLINK_NAMESPACE, 'href', node.attrs.href)
-        linkNode.textContent = text
+        const $linkNode = this.createElement('ext-link')
```
mnatsheh: Maybe `$extLink`.

src/jats/exporter/jats-exporter.ts
```diff
@@ -1767,340 +1894,108 @@ export class JATSExporter {
...
-    return transAbstractNode
-  }
-
-  private createAbstractNode(abstractSection: Element): Element {
-    const abstractNode = this.createElement('abstract')
-    for (const node of abstractSection.childNodes) {
-      if (node.nodeName !== 'title') {
-        abstractNode.appendChild(node.cloneNode(true))
+  private buildAbstracts = (): Element[] => {
+    const abstracts: Element[] = []
```
mnatsheh: `$abstracts`?

src/jats/exporter/jats-exporter.ts
```diff
@@ -1767,340 +1894,108 @@ export class JATSExporter {
...
-    footNote.setAttribute('fn-type', fnType)
-    const title = section.querySelector('title')
-    if (title) {
-      const footNoteTitle = this.createElement('label')
-      footNoteTitle.textContent = title.textContent
-      section.removeChild(title)
-      footNote.append(footNoteTitle)
-    }
-    footNote.append(...section.children)
+    const $footNote = this.createElement('fn')
```
mnatsheh: `$fn`?

src/jats/importer/jats-transformations.ts
mnatsheh: Is this to handle the slight discrepancy between the schema and jats? if so, we can undo that change to the schema.

mnatsheh (commented):
I've skimmed through this, and marked a few things. I'll take another look in the next few days.

I've also asked Claude to review and it reported the following:
> 1. jats-exporter.ts:224 — extractedSections is populated from routing.extracted after buildBack() already ran, so back-matter serialization sees a stale empty set. A back-matter section nested inside another back-matter section gets rendered twice in
  the output — once embedded in its parent, once as its own independently-extracted entry. This directly contradicts the code's own comment.
> 2. jats-exporter.ts:614 — classify() returns early for a second section with a "unique" category (availability/ethics-statement/acknowledgements) before marking it extracted. The duplicate silently leaks into <body> as a plain <sec> instead of being
  suppressed/routed.
> 3. jats-dom-parser.ts:1100 — the new abstract tag's parseDOM rule only resolves category via chooseSectionCategory() when an abstract-type attribute is present; otherwise hardcodes ''. Plain <abstract> elements (the common main-abstract case) now
  import with an empty category instead of "abstract" — confirmed by the snapshot diff. Every sibling trans-abstract rule calls this unconditionally, so this one looks like an oversight.

src/jats/importer/jats-transformations.ts
mbetamony: Do you mean undoing the new abstracts node? Isn't it better to have the stricter schema? I don't see what issues this sort causes to be honest, it just ensures we have the right sections in the right order.

mnatsheh: No. In JATS, the article-meta schema has the following `..., (abstract)*, trans-abstract*, ...`. This means that we are guaranteed to get regular abstracts before translated ones, but there is no distinction between text and graphical ones. 

This means that if we change our abstracts node content schema back to be `abstract* trans_abstract*`, we should not need to do any sorting.

mbetamony:
I've updated all the review points (Hopefully i didn't miss any). 

I also added an ignore rule for `statement` element in the dom-parser, it used to be parsed a section which i wasn't sure if that was correct. 

I also simplified the backmatter section routing.

asouqi (approved):

